PR #4, opened 2021-06-09
[Feature] Employee GraphQL queries

OustanDing:
- Created Employee GraphQL queries and mutations (create)
- Added enums to top-level GraphQL schema
- Isolated GraphQL `context` into separate file
- Small fix to `generate-graphql-types` script (was breaking if `types.ts` did not exist)
- Removed unused configuration in `apollo.config.js`

### DIFF
--- a/lib/employees/resolvers.ts
+++ b/lib/employees/resolvers.ts
@@ -1,5 +1,7 @@
-/*eslint-disable */
+import { ApolloError } from 'apollo-server-errors'; // Apollo error
 import { Resolver } from '@lib/resolvers'; // Resolver type
+import { EmployeeAlreadyExistsError } from '@lib/employees/errors'; // Employee errors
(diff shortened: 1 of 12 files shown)


lib/employees/resolvers.ts
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Create an RCD employee
+ * @returns Status of operation (ok, error)
+ */
+export const createEmployee: Resolver = async (_, args, { prisma }) => {
+  const {
+    input: { firstName, lastName, email, role },
+  } = args;
+
+  await prisma.employee.create({
```
angeladietz, line 22: Do we need to be surrounding this `prisma.employee.create` in a try-catch? If not, what happens if the creation fails for some reason?

angeladietz (approved):
Looks great 🚀

PR #6, opened 2021-06-12
[Feature] Set up Next Auth authentication

OustanDing:
- Set up Next Auth with email provider
- Create `sendVerificationRequest` function for sending email to users
- Update DB and Prisma schema to store verification requests and email verifications
- Created basic login page for testing (temporary)
- Disallow registration (only allow login for accounts that exist in the DB)


pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,99 @@
...
+      server: {
+        host: process.env.NA_EMAIL_HOST as string,
+        port: parseInt(process.env.NA_EMAIL_PORT as string, 10),
+        auth: {
+          user: process.env.NA_EMAIL_USER as string,
+          pass: process.env.NA_EMAIL_PASSWORD as string,
+        },
+      },
+      from: process.env.NA_EMAIL_FROM,
+      maxAge: 24 * 60, // 1 hour max life for login request
```
Comment (jeffreyzhang2001) [inline pages/api/auth/[...nextauth].ts:39]: Should be 60 * 60 (1 hour)

pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,99 @@
...
+      return Promise.resolve(session);
+    },
+    jwt: async (token, user) => {
+      user ? (token.role = user.role) : null;
+      user ? (token.id = user.id) : null;
+
+      return Promise.resolve(token);
+    },
+    signIn: async user => {
+      if (!user || !user.email) {
```
Comment (jeffreyzhang2001) [inline pages/api/auth/[...nextauth].ts:78]: I think we can just do !user?.email here.

jeffreyzhang2001 (changes requested):
LGTM from some local testing (other than some small comments). Would be nice to have a logout button as well before merging it in. Good work :)

pages/api/auth/[...nextauth].ts
Comment (OustanDing) [reply pages/api/auth/[...nextauth].ts]: Good catch

jeffreyzhang2001:
LGTM!

jeffreyzhang2001 (approved):

PR #7, opened 2021-06-12
[Feature] Graphql endpoints for the physician model

EmilioMena:
- Added physician queries and mutations

I deviated a bit from the error handling section that's [on notion](https://www.notion.so/uwblueprintexecs/Development-processes-a0313f2bef9941fc8752c8e7b5be5e70#9e88e7b5ccb44f798d09b665845092a1) for a couple reasons which I commented on. Still not 100% sure it's the best way to go in this case so would love some feedback on that

Mutation:
image

Query
image

Errors:
Duplicate MSP
image
Length Constraint
image

### DIFF
--- a/lib/graphql/schema.graphql
+++ b/lib/graphql/schema.graphql
@@ -2,11 +2,13 @@ type Query {
   meta: Meta!
   applicants: [Applicant!]
   employees: [Employee!]
+  physicians: [Physician!]
 }
 
 type Mutation {
   createApplicant(input: CreateApplicantInput!): CreateApplicantResult!
   createEmployee(input: CreateEmployeeInput!): CreateEmployeeResult!
+  createPhysician(input: CreatePhysicianInput!): CreatePhysicianResult!
 }
 
 scalar Date
--- a/lib/physicians/errors.ts
+++ b/lib/physicians/errors.ts
@@ -1,0 +1,12 @@
+import { ApolloError } from 'apollo-server-errors';
+
+/**
+ * Physician create errors
+ */
+export class PhysicianCreateError extends ApolloError {
+  constructor(message: string) {
+    super(message, 'PHYSICIAN_CREATE_ERROR');
+
+    Object.defineProperty(this, 'name', { value: 'PhysicianCreateError' });
+  }
+}
--- a/lib/physicians/resolvers.ts
+++ b/lib/physicians/resolvers.ts
@@ -1,0 +1,51 @@
+import { ApolloError } from 'apollo-server-errors'; // Apollo error
+import { Resolver } from '@lib/resolvers'; // Resolver type
+import { DBErrorCode } from '@lib/db/errors'; // Database errors
+import { PhysicianCreateError } from '@lib/physicians/errors'; // Physician error
+import { Prisma } from '.prisma/client';
+
+/**
+ * Query all physicians of RCD applicants
+ * @returns All physicians of RCD applicants
+ */
+export const physicians: Resolver = async (_parent, _args, { prisma }) => {
+  const physicians = await prisma.physician.findMany();
+  return physicians;
+};
+
+/**
+ * Create a physician
+ * @returns Status of operation (ok)
+ */
+export const createPhysician: Resolver = async (_, args, { prisma }) => {
+  const { input } = args;
+
+  let physician;
+  try {
+    physician = await prisma.physician.create({
+      data: { ...input },
+    });
+  } catch (err) {
+    if (err instanceof Prisma.PrismaClientKnownRequestError) {
+      if (err.code === DBErrorCode.UniqueConstraintFailed) {
+        throw new PhysicianCreateError(
+          'Unique constraint failed, a new physician cannot be created with this MSP number'
+        );
+      }
+      if (err.code === DBErrorCode.LengthConstraintFailed) {
+        throw new PhysicianCreateError(
+          "Length constraint failed, provided value too long for the column's type"
+        );
+      }
+    }
+  }
+
+  // Throw internal server error if physician was not created
+  if (!physician) {
+    throw new ApolloError('Physician was unable to be created');
+  }
+
+  return {
+    ok: true,
+  };
+};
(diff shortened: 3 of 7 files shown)


lib/physicians/resolvers.ts
```diff
@@ -0,0 +1,75 @@
...
+    if (err instanceof Prisma.PrismaClientKnownRequestError) {
+      if (err.code === DBErrorCode.UniqueConstraintFailed) {
+        throw new PhysicianError(
+          'Unique constraint failed, a new physician cannot be created with this MSP number'
+        );
+      }
+      if (err.code === DBErrorCode.LengthConstraintFailed) {
+        throw new PhysicianError(
+          "Length constraint failed, provided value too long for the column's type"
+        );
```
EmilioMena, line 62: I handled each error just by code and not by the column since the [LengthConstraintError](https://www.prisma.io/docs/reference/api-reference/error-reference#p2000) didn't actually return which column was failing in the `meta `property for some reason. 
image

Also, there are a lot of fields that have this length constraint on them so making an error object + adding an extra conditional for each one would start getting pretty ugly imo. 

Don't feel too strongly about this tho so definitely open to other ideas

lib/physicians/resolvers.ts
```diff
@@ -0,0 +1,75 @@
...
+        city,
+        province,
+        postalCode,
+        phone,
+        status,
+        notes,
+      },
+    });
+  } catch (err) {
+    if (err instanceof Prisma.PrismaClientKnownRequestError) {
```
EmilioMena, line 53: I also added this to make sure that the error is a prisma query engine error so that `err.code` wouldn't fail

lib/graphql/schema.graphql
```diff
@@ -1,12 +1,17 @@
...
+  physicians: [Physician!]
 }
 
 type Mutation {
   createEmployee(input: CreateEmployeeInput!): CreateEmployeeResult!
 }
 
+type Mutation {
+  createPhysician(input: CreatePhysicianInput!): CreatePhysicianResult!
+}
```
OustanDing, line 13: I think this should go under the `type Mutation` above unless that doesn't work

lib/physicians/errors.ts
```diff
@@ -0,0 +1,12 @@
+import { ApolloError } from 'apollo-server-errors';
+
+/**
+ * Physician create errors
+ */
+export class PhysicianError extends ApolloError {
```
OustanDing, line 6:
```suggestion
export class PhysicianCreateError extends ApolloError {
```

lib/physicians/errors.ts
```diff
@@ -0,0 +1,12 @@
...
+  constructor(message: string) {
+    super(message, 'PHYSICIAN_CREATE_ERROR');
+
+    Object.defineProperty(this, 'name', { value: 'PhysicianError' });
```
OustanDing, line 10:
```suggestion
    Object.defineProperty(this, 'name', { value: 'PhysicianCreateError' });
```

lib/physicians/resolvers.ts
```diff
@@ -0,0 +1,75 @@
...
+ * @returns All physicians of RCD applicants
+ */
+export const physicians: Resolver = async (_parent, _args, { prisma }) => {
+  const physicians = await prisma.physician.findMany();
+  return physicians;
+};
+
+/**
+ * Create a physician
+ * @returns Status of operation (ok, error)
```
OustanDing, line 18:
```suggestion
 * @returns Status of operation (ok)
```

lib/physicians/resolvers.ts
```diff
@@ -0,0 +1,75 @@
...
+      lastName,
+      mspNumber,
+      address,
+      city,
+      province,
+      postalCode,
+      phone,
+      status,
+      notes,
+    },
```
OustanDing, line 33:
```suggestion
    input,
```
Could just pull out `input` here and destructure it below

lib/physicians/resolvers.ts
```diff
@@ -0,0 +1,75 @@
...
+        lastName,
+        mspNumber,
+        address,
+        city,
+        province,
+        postalCode,
+        phone,
+        status,
+        notes,
+      },
```
OustanDing, line 50:
```suggestion
      data: { ...input },
```

lib/physicians/resolvers.ts
OustanDing: Good catch

lib/physicians/resolvers.ts
```diff
@@ -0,0 +1,75 @@
...
+    });
+  } catch (err) {
+    if (err instanceof Prisma.PrismaClientKnownRequestError) {
+      if (err.code === DBErrorCode.UniqueConstraintFailed) {
```
OustanDing, line 54: Also want to check that the field that caused this error was the MSP number

lib/physicians/resolvers.ts
OustanDing: I see. Sounds good. In the future, let's investigate and see if we can include the field that failed the length constraint in the error message.

OustanDing (changes requested):

EmilioMena:
🏓 @OustanDing added those changes - ready for another look

lib/physicians/resolvers.ts
```diff
@@ -0,0 +1,51 @@
+import { ApolloError } from 'apollo-server-errors'; // Apollo error
+import { Resolver } from '@lib/resolvers'; // Resolver type
+import { DBErrorCode } from '@lib/db/errors'; // Database errors
+import { PhysicianCreateError } from './errors'; // Physician error
+import { Prisma } from '.prisma/client';
```
OustanDing, line 5:
```suggestion
import { PhysicianCreateError } from '@lib/physicians/errors'; // Physician error
import { Prisma } from '@prisma/client';
```

Avoid relative imports

OustanDing (changes requested):
One small thing

lib/physicians/resolvers.ts
EmilioMena: good call, missed that

OustanDing (approved):
LGTM!

PR #11, opened 2021-06-16
[Feature] GraphQL queries for permits model

jennifertsai:
## Notion ticket link

[Set up GQL queries for Permits](https://www.notion.so/uwblueprintexecs/Set-up-GQL-queries-for-Permits-a163b57945524a97bb143d08c1bdd41f)

## Implementation description
- Added GraphQL queries `permits` and mutations `createPermit`
- Generated Typescript types
- BLOCKED: referencing application and applicant from `applicationId` and `applicantId`

## Checklist

### DIFF
--- a/lib/permits/resolvers.ts
+++ b/lib/permits/resolvers.ts
@@ -1,6 +1,10 @@
 import { ApolloError } from 'apollo-server-errors'; // Apollo error
 import { Resolver } from '@lib/resolvers'; // Resolver type
-import { PermitAlreadyExistsError } from '@lib/permits/errors'; // Employee errors
+import {
+  PermitAlreadyExistsError,
+  ApplicantIdDoesNotExistError,
+  ApplicationIdDoesNotExistError,
+} from '@lib/permits/errors'; // Employee errors
 import { DBErrorCode } from '@lib/db/errors'; // Database errors
 
 /**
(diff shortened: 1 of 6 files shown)


prisma/schema.prisma
```diff
@@ -190,23 +190,23 @@ enum Aid {
   SCOOTER
   WALKER
 
-  @@map("aid")
+   @@map("aid")
```
OustanDing, line 193: Not sure why these spaces are being appended. Can you run `prisma format` before committing? That should prevent these spaces from being added.

OustanDing (changes requested):
Just need to fix the formatting in the Prisma schema file. Also, please rebase on `staging` to pull in the Next.js linting fixes so that the build can succeed!

lib/permits/resolvers.ts
```diff
@@ -0,0 +1,53 @@
...
+  try {
+    permit = await prisma.permit.create({
+      data: {
+        rcdPermitId,
+        expiryDate,
+        receiptId,
+        active,
+        applicationId,
+        applicantId,
+      },
```
EmilioMena, line 34: if you pull out input above, you can unpack it here instead to save a few lines:
```suggestion
      data: { ...input },
```

lib/permits/resolvers.ts
```diff
@@ -0,0 +1,53 @@
...
+};
+
+/**
+ * Create an RCD permit
+ * @returns Status of operation (ok, error)
+ */
+export const createPermit: Resolver = async (_, args, { prisma }) => {
+  const {
+    input: { rcdPermitId, expiryDate, receiptId, active, applicationId, applicantId },
+  } = args;
```
EmilioMena, line 22: a neat trick i learned from Oustan is that you can take out input instead and unpack the arguments when creating the record. There's an example [here ](https://github.com/uwblueprint/richmond-centre-for-disability/blob/staging/lib/physicians/resolvers.ts)
```suggestion
  const { input } = args;
```

lib/permits/resolvers.ts
```diff
@@ -0,0 +1,53 @@
...
+      data: {
+        rcdPermitId,
+        expiryDate,
+        receiptId,
+        active,
+        applicationId,
+        applicantId,
+      },
+    });
+  } catch (err) {
```
EmilioMena, line 36: There should probably be foreign key constraint error handling for application_id and applicant_id here as well since there can't be a permit without an application and applicant. I think angela added the error code in #12 so maybe wait for that to merge to add that in?

EmilioMena (approved):
Looks good! 🚀

OustanDing (approved):
LGTM!

PR #16, opened 2021-06-20
[Feature] Add formatting functions for phone number and postal code

EmilioMena:
## Notion ticket link

[Function for formatting db fields](https://www.notion.so/uwblueprintexecs/Function-for-formatting-DB-fields-887ec602da7043539495d87bfa44871f)

## Implementation description
* added formatting functions for phone number and postal codes under `lib/utils/format.ts` which remove extra characters from these fields
* called these functions in applicant and physician resolvers

## Notes
* Some sort of data validation should probably be done before formatting, didn't put it in this PR to keep things concise 
* any other formatting that should be done before saving?

## Checklist

### DIFF
--- a/lib/applicants/resolvers.ts
+++ b/lib/applicants/resolvers.ts
@@ -2,6 +2,7 @@ import { ApolloError } from 'apollo-server-errors'; // Apollo error
 import { Resolver } from '@lib/resolvers'; // Resolver type
 import { ApplicantAlreadyExistsError } from '@lib/applicants/errors'; // Employee errors
 import { DBErrorCode } from '@lib/db/errors'; // Database errors
+import { formatPhoneNumber, formatPostalCode } from '@lib/utils/format';
 
 /**
  * Query all the RCD applicants in the internal-facing app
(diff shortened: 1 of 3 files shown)


lib/utils/format.ts
```diff
@@ -0,0 +1,20 @@
+enum FormatRegex {
+  PHONE_NUMBER = '/D/g',
+  POSTAL_CODE = '/[^A-Za-zd]/g',
```
EmilioMena, line 3: the postal code regex is supposed to be 
```suggestion
  POSTAL_CODE = '/[^A-Za-z\d]/g',
```
but the precommit linter removes the `\` - anyone know how to get around it?

OustanDing: I don't think the regex patterns are supposed to be stored as strings (eg. we would call `string.replace(/[^A-Za-zd]/g)` instead of `string.replace('/[^A-Za-zd]/g')`. The most straightforward solution would be to put the regex patterns directly in the formatting functions rather than having this enum.

lib/applicants/resolvers.ts
```diff
@@ -17,18 +18,18 @@ export const applicants: Resolver = async (_parent, _args, { prisma }) => {
...
-  } = args;
+  const { input } = args;
 
   let applicant;
   try {
+    input.postalCode = formatPostalCode(input.postalCode);
+    input.phone = formatPhoneNumber(input.phone);
     applicant = await prisma.applicant.create({
       data: { ...args.input },
     });
```
OustanDing, line 29: We don't want to modify the input args to this function as it could cause debugging issues down the road and is generally bad practice. Instead, we could do something like this to format the fields without modifying the function parameters:
```suggestion
    applicant = await prisma.applicant.create({
      data: { 
          ...args.input,
          postalCode: formatPostalCode(input.postalCode),
          phone: formatPhoneNumber(input.phone),
       },
    });
```

lib/physicians/resolvers.ts
```diff
@@ -22,6 +23,8 @@ export const createPhysician: Resolver = async (_, args, { prisma }) => {
 
   let physician;
   try {
+    input.postalCode = formatPostalCode(input.postalCode);
+    input.phone = formatPhoneNumber(input.phone);
```
OustanDing, line 27: Similar issue to that in `lib/applicants/resolvers.ts`

OustanDing (changes requested):
Just a small comment regarding best practices

lib/applicants/resolvers.ts
EmilioMena: good point, I like this a lot better

OustanDing (approved):
LGTM!

PR #19, opened 2021-06-30
[Feature] Add auth wrapper to graphql resolvers

EmilioMena:
## Notion ticket link

[Protect graphql routes](https://www.notion.so/uwblueprintexecs/Decorator-protecting-authenticated-GraphQL-routes-034cab918d2c4e87addd891d4c9e6ff2)

## Implementation description
* Added session to graphQL context
* Added an authorization wrapper function that protects graphQL routes based on an employees's role that is found in the session. The wrapper accepts a resolver function and an array of roles that can access it.
* wrapped all resolvers in this function

## Notes
* Since admins should have access to all routes, instead of passing Role.admin every time we could just check that in the wrapper.
* Couldn't figure out how to get the JWT after logging in so wasn't able to test as an authorized user. 

## Checklist


lib/graphql/auth-guards.ts
```diff
@@ -0,0 +1,16 @@
+import { Role } from '@prisma/client';
+import { Resolver } from './resolvers';
+
+export const authenticate =
+  (resolver: Resolver, authorizedRoles: Array<Role>): Resolver =>
+  (root, args, context, info) => {
+    const userRole = context.session?.user?.role;
```
EmilioMena, line 7: build is failing with error `'role' does not exist` on the user. Kinda confused by this since i thought that get_session would return what is included in the next_auth session callback.
https://github.com/uwblueprint/richmond-centre-for-disability/blob/06c6b000f9505440d08e18a26464cbb960d5e3c8/pages/api/auth/%5B...nextauth%5D.ts#L65-L71

Alternative would be to make a db call with the session email in context.ts and get the user that way. Anyone have any thoughts?

OustanDing: Hmm, looks like it's an issue with GraphQL playground configuration. This SO post has a solution https://stackoverflow.com/a/66308267.

lib/graphql/apollo.ts
```diff
@@ -12,5 +12,5 @@ export const apolloServer = new ApolloServer({
       'schema.polling.enable': false, // Disable infinite schema introspection
     },
   },
-  context,
+  context: context,
```
OustanDing, line 15: This can be left as it was before

lib/graphql/auth-guards.ts
```diff
@@ -0,0 +1,16 @@
+import { Role } from '@prisma/client';
+import { Resolver } from './resolvers';
```
OustanDing, line 2:
```suggestion
import { Resolver } from '@lib/resolvers';
```

lib/graphql/auth-guards.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const authenticate =
+  (resolver: Resolver, authorizedRoles: Array<Role>): Resolver =>
```
OustanDing, line 5:
```suggestion
  (resolver: Resolver, authorizedRoles: ReadonlyArray<Role>): Resolver =>
```

lib/graphql/auth-guards.ts
```diff
@@ -0,0 +1,16 @@
+import { Role } from '@prisma/client';
+import { Resolver } from './resolvers';
+
+export const authenticate =
```
OustanDing, line 4: Please add some documentation 😄

lib/graphql/context.ts
```diff
@@ -1,12 +1,19 @@
...
+import { Session } from 'next-auth';
+import { getSession } from 'next-auth/client';
 
 export type Context = {
   prisma: PrismaClient;
+  session: Session | null;
 };
 
-export default function context(): Context {
+export default async function context(req: MicroRequest): Promise<Context> {
```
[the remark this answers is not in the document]
OustanDing: Try this first before adding `request.credentials: include` to our `ApolloServer` configuration

lib/graphql/resolvers.ts
```diff
@@ -7,25 +7,27 @@ import { permits, createPermit } from '@lib/permits/resolvers';
...
 
 const resolvers = {
   Query: {
-    meta,
-    applicants,
-    employees,
-    physicians,
-    applications,
-    permits,
+    meta: meta,
```
OustanDing, line 18:
```suggestion
    meta,
```
Small nit but I think it reduces redundancy.

lib/graphql/resolvers.ts
```diff
@@ -7,25 +7,27 @@ import { permits, createPermit } from '@lib/permits/resolvers';
 import { IFieldResolver } from 'graphql-tools'; // GraphQL field resolver
 import { Context } from '@lib/context'; // Context type
 import { dateScalar } from '@lib/scalars'; // Custom date scalar implementation
+import { authenticate } from '@lib/auth-guards';
+import { Role } from '@lib/graphql/types';
 
 // Resolver type
 export type Resolver<P = undefined> = IFieldResolver<P, Context>;
 
 const resolvers = {
```
OustanDing, line 16: Would be great if you could add a small comment above this line briefly explaining the usage of `authenticate` here to protect the routes and restrict them to RCD employee roles.

lib/graphql/auth-guards.ts
```diff
@@ -0,0 +1,16 @@
+import { Role } from '@prisma/client';
```
OustanDing, line 1: Maybe change this file name to just `authorization.ts`. Thoughts?

lib/graphql/auth-guards.ts
OustanDing:
```suggestion
export const authorize =
```
Authentication means login. Authorization means checking for roles and access to certain features, so I think this should be named `authorize`.

lib/graphql/auth-guards.ts
```diff
@@ -0,0 +1,16 @@
...
+import { Resolver } from './resolvers';
+
+export const authenticate =
+  (resolver: Resolver, authorizedRoles: Array<Role>): Resolver =>
+  (root, args, context, info) => {
+    const userRole = context.session?.user?.role;
+
+    // If user has the necessary permissions, allow the query or mutation to go through.
+    if (userRole && authorizedRoles.includes(userRole)) {
```
OustanDing, line 10: We also want to give admins full access to the platform here

lib/graphql/auth-guards.ts
EmilioMena: nice catch, had it like this since i was originally planning on authenticating in the wrapper and not in context

lib/graphql/auth-guards.ts
EmilioMena: Role.Admin is passed in every resolver that's wrapped so admins will always have access. I guess to remove redundancy we could just add an extra check here for admins instead of always passing them in as authorized roles. What do you think?

lib/graphql/auth-guards.ts
EmilioMena: tried this but didn't do the trick. After a bunch of looking around i found that we have to override the default next-auth types if we change what they output https://next-auth.js.org/getting-started/typescript#module-augmentation

lib/graphql/auth-guards.ts
OustanDing: Yeah, exactly what I was thinking!

lib/graphql/authorization.ts
```diff
@@ -0,0 +1,22 @@
+import { Role } from '@prisma/client';
```
OustanDing, line 1: Small nit, but for consistency, let's use `@lib/types` here.

types/next-auth.d.ts
```diff
@@ -0,0 +1,17 @@
+import NextAuth from 'next-auth'; // eslint-disable-line @typescript-eslint/no-unused-vars
```
OustanDing, line 1: Just wondering where this file came from/what it's used for?

OustanDing (commented):
Sorry for the late review! It would be great if we could include a separate check for `ADMIN` in the `authorize` function. Also, two small other comments :)

types/next-auth.d.ts
EmilioMena: this is from the docs, I wanted to remove it but typescript complains if it's not there for some reason 
https://next-auth.js.org/getting-started/typescript#:~:text=ts-,import%20nextauth%20from%20%22next-auth%22,-declare

EmilioMena:
🏓 @OustanDing added the changes you requested - ready for another review!

types/next-auth.d.ts
OustanDing: Got it!

OustanDing (approved):
LGTM! Thanks for the changes :D

PR #20, opened 2021-07-01
[Feature] Updated login email template

OustanDing:
## Notion ticket link

[Ticket](https://www.notion.so/uwblueprintexecs/Build-Login-verification-email-3a77cdfdaf3b45f0ae0e824d115fb774)

## Implementation description
* Updated the login verification email appearance in accordance with hi-fi designs

## Notes
* The image for the email is hard-coded, so PR needs to be merged into staging before it will properly display in the email. We also need to change the image source when we make our initial production releases to point to an image source in production.

## Checklist


lib/auth/sendVerificationRequest.ts
```diff
@@ -45,9 +45,86 @@ const html = (config: EmailConfig) => {
...
+          <p
+            style="
+              font-size: 18px;
+              line-height: 150%;
+              margin: 0 0 32px;
+            "
+          >
+            Click the button below to authenticate as ${escapedEmail}
+            <br />
+            and sign into RCD APP Management System.
```
EmilioMena, line 99: just thinking out loud but it might be handy to make the "Platform name" a constant somewhere just so if we or RCD ever want it changed we don't have to go around looking everywhere

lib/auth/sendVerificationRequest.ts
```diff
@@ -45,9 +45,86 @@ const html = (config: EmailConfig) => {
...
+              background-color: #0B64CC;
+              color: #ffffff;
+              padding: 12px 24px;
+              border-radius: 6px;
+              font-size: 18px;
+              line-height: 150%;
+              text-decoration: none;
+            "
+          >
+            Log into RCD APP Management System
```
EmilioMena, line 113: i think this needs to be semi-bold from the design, could be wrong

lib/auth/sendVerificationRequest.ts
```diff
@@ -45,9 +45,86 @@ const html = (config: EmailConfig) => {
   const escapedEmail = `${email.replace(/\./g, '&#8203;.')}`;
 
   // HTML email template
```
EmilioMena, line 47: email subject needs to be changed to 
> Your temporary APP Manager login

EmilioMena (commented):
Nice work! Left a couple small comments

lib/auth/sendVerificationRequest.ts
OustanDing: Good point. However, the ideal solution is to confirm the platform name early so that we don't change it at all. It shouldn't be something that changes often, and the name of the platform isn't displayed a lot on the platform itself, so it might be a little overkill for our case. I will confirm the internal platform name with Carelynn and RCD ASAP.

lib/auth/sendVerificationRequest.ts
OustanDing: Good point. Thanks for the catch!

lib/auth/sendVerificationRequest.ts
OustanDing: Updated the name!

EmilioMena (approved):
Thanks for those changes looks good! 🚢

PR #22, opened 2021-07-07
[Feature] Create applicant status badge component

j-bunk:
## Notion ticket link

[Build ApplicationStatusBadge component](https://www.notion.so/uwblueprintexecs/Build-ApplicationStatusBadge-component-00e9b596661a439b8cfb352e1961f0fb)

## Implementation description
* [Figma designs](https://www.figma.com/file/0J04jrcDl6CiR24IfyjjF8/RCD-%2F-Design-System?node-id=37%3A0) for the badges can be found in the badges section to the right
* Modified Chakra UI's badge theming in `tools/theme/components/badge.ts` and then created pre-built badges for the Applicant's status in `components/internal/ApplicantStatusBadge.tsx` 
* I exported the badge icons as svg from the Figma designs

Example of how to use the badges:
```jsx
<Stack direction="row">
          {/*  Normal badges, variant specifies the color */}
          <Badge variant="active">Active</Badge>
          <Badge variant="inactive">Inactive</Badge>
          {/*Applicant status badge, they are prebuilt and no text needs to be written */}
          <ApplicantStatusBadge variant="expiring" />
          <ApplicantStatusBadge variant="completed" />
          <ApplicantStatusBadge variant="inProgress" />
          <ApplicantStatusBadge variant="pending" />
          <ApplicantStatusBadge variant="rejected" />
          <ApplicantStatusBadge variant="expired" />
          <ApplicantStatusBadge variant="active" />
</Stack>
```

## Image of all the badges
image

## Notes
* I'm not the strongest frontend developer so please nitpick and lmk if I'm doing anything wrong
* Resolved! ~~I couldn't get the exact positioning /spacing right. There's a bit too much space to the left of the badge icon, if someone's nifty with frontend/CSS and knows how to fix it lmk.~~
### Badge created
image
### Figma Badge
image

## Checklist


tools/theme/components/badge.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  // Variants (completed, inProgress, pending, rejected, expiring, expired, active, inactive)
+  variants: {
+    completed: () => {
+      return {
+        bg: '#EEEEEE',
+      };
+    },
+    inProgress: () => ({
+      bg: '#D9F1F8',
```
j-bunk, line 26: Is it ok to specify the colors like this? They don't exist in `tools/theme/colors.ts`

chrischan325 (commented):
Looks good to me! Can't really speak on the specifics of the styling but everything else looks good!

tools/theme/components/badge.ts
angeladietz: In my opinion I think it would be better to add these colours to that `theme/colors.ts, especially since some of them are used multiple times throughout this file

components/internal/ApplicantStatusBadge.tsx
```diff
@@ -0,0 +1,118 @@
+import { Badge, Box } from '@chakra-ui/react'; // Chakra UI
+import Image from 'next/image'; // Optimized images
+
+type Props = {
+  variant: string;
```
OustanDing, line 5: We can make this a union of strings:
`variant: "completed" | "expiring" | "rejected" | ...`
to enforce the allowed variants

components/internal/ApplicantStatusBadge.tsx
```diff
@@ -0,0 +1,118 @@
...
+};
+
+/**
+ * Renders the applicant status badge content (icon + text)
+ * @param variant for specifying applicant status badge type
+ * @returns {JSX.Element} contents of the badge
+ */
+const renderBadgeContent = (variant: string): JSX.Element | undefined => {
```
OustanDing, line 13: We don't want to render `undefined` - we can work around this with the suggestion I posted above. Then, we can replace the `switch` statement with an `if-else` statement and cover all of the possible values. TypeScript shouldn't complain if we do that.

components/internal/ApplicantStatusBadge.tsx
```diff
@@ -0,0 +1,118 @@
...
+          <Box as="span" ml="2">
+            Active
+          </Box>
+        </Box>
+      );
+  }
+};
+
+export function ApplicantStatusBadge({ variant }: Props) {
+  return <Badge variant={variant}>{renderBadgeContent(variant)}</Badge>;
```
OustanDing, line 117: In the `switch` statement you have above, the outside `Box` component is the same for all of the variants, so we can include it in the `return` statement here. Then, we want the returned JSX to be wrapped in a React fragment, like this:
```javascript
if (variant === 'active') {
  return (
    <>
      <Image />
      <Box>Active</Box>
    </>
  )
}
```

Also, after that, we want to move the `renderBadgeContent` function to be within the scope of the `ApplicantStatusBadge` functional component and rename it to `_renderBadgeContent` as it shouldn't be used elsewhere.

tools/theme/components/badge.ts
```diff
@@ -0,0 +1,49 @@
...
+  // Base style
+  baseStyle: {
+    display: 'flex',
+    alignItems: 'center',
+    borderRadius: '23px',
+    textTransform: 'inherit',
+    fontWeight: 'normal',
+    fontSize: '16px',
+    height: '32px',
+    paddingX: '1rem',
```
OustanDing, line 14: I think that we can adjust the left padding a bit here to address the issue of the left part of the badge being too wide.

tools/theme/components/badge.ts
OustanDing: ^ agree with @angeladietz here. We should add a `badge` property to the `colors` object that contains all of these colours, which we can reference here.

tools/theme/components/badge.ts
```diff
@@ -0,0 +1,49 @@
...
+    paddingY: '4px',
+  },
+
+  // Variants (completed, inProgress, pending, rejected, expiring, expired, active, inactive)
+  variants: {
```
OustanDing, line 19: Since we're just returning props objects here, we can replace all of the functions with plain objects, like so:
```javascript
completed: {
  bg: '#EEEEEE',
}
```

OustanDing (changes requested):
Looks great so far! Made some suggestions that will hopefully make the code a bit cleaner/more reusable :)

tools/theme/components/badge.ts
j-bunk: Adjusting it here messed up the positioning for the badges without an icon so I ended up adjusting it in `components/internal/ApplicantStatusBadge.tsx` by changing the left margin value here
```tsx
return (
    <Badge variant={variant}>
      <Box ml="-1" mr="2">
        {_renderBadgeContent(variant)}
      </Box>
    </Badge>
  );
  ```

File: tools/theme/components/badge.ts (hunk 1 above)
Comment (j-bunk) [reply tools/theme/components/badge.ts]: Done! Looks much cleaner  🧹

File: components/internal/ApplicantStatusBadge.tsx (hunk 3 above)
Comment (j-bunk) [reply components/internal/ApplicantStatusBadge.tsx]: I tried that out and it still complains unless I have a final `else` statement but that's the equivalent of specifying the `default` case in a switch statement. I opted to still use the switch statment but made the `active` case the default case so that fixes it. Lmk if you have any issues with that!

Code snippet:
```ts
case 'active':
      default:
        return (
....
```

j-bunk:
@OustanDing Addressed the feedback and it looks a lot cleaner 🧹 It's ready for another review!

components/internal/ApplicantStatusBadge.tsx
```diff
@@ -0,0 +1,133 @@
...
+type Props = {
+  variant:
+    | 'completed'
+    | 'inProgress'
+    | 'pending'
+    | 'rejected'
+    | 'expiring'
+    | 'expired'
+    | 'active'
+    | 'inactive';
```
OustanDing, line 13: I think we're missing the `inactive` variant in the `switch` below?

components/internal/ApplicantStatusBadge.tsx
OustanDing: That is fine!

components/internal/ApplicantStatusBadge.tsx
```diff
@@ -0,0 +1,133 @@
...
+              Active
+            </Box>
+          </>
+        );
+    }
+  };
+
+  return (
+    <Badge variant={variant}>
+      <Box ml="-1" mr="2">
```
OustanDing, line 128: Are these pixels? If yes, could you please include the "px"? Just for consistency

tools/theme/components/badge.ts
OustanDing: Got it

tools/theme/components/badge.ts
```diff
@@ -0,0 +1,47 @@
...
+      bg: 'badge.pending',
+    },
+    rejected: {
+      bg: 'badge.rejected',
+    },
+    expiring: {
+      bg: 'badge.expiring',
+    },
+    expired: {
+      bg: 'badge.rejected',
```
OustanDing, line 36: Should this be expired?

j-bunk: The expired color and rejected color are the same so it's being reused here. If that's confusing I could rename the color with something that tries to encompass both rejected and expired, or restate the color again in `tools/theme/colors.ts` with the name expired

components/internal/ApplicantStatusBadge.tsx
j-bunk: The inactive badge doesn't have an icon and so you would use a normal badge to create it
```tsx
<Badge variant="inactive">Inactive</Badge>
```
But I should remove `inactive` from the union of strings (and this explains why a final else/default was required, if I remove it I no longer need to specify that).

tools/theme/components/badge.ts
OustanDing: I think to reduce confusion, we should just have the two colours in `colors.ts`. There will be a bit of repeated "code", but it is clearer and reduces potential confusion.

components/internal/ApplicantStatusBadge.tsx
j-bunk: I believe this follows the spacing in `tools/theme/spacing.ts` which makes it equivalent to 
```tsx
      <Box ml="-4px" mr="8px">
```
Specifying the pixels seems more readable so I'll it to that instead

components/internal/ApplicantStatusBadge.tsx
OustanDing: Hm, we have all of the other variants encompassed in this `ApplicantStatusBadge` component, so I feel like we should include the `inactive` variant as well, even though it is trivial to render it using the `Badge` component. This gives us consistency and a single source of truth for the rendering of applicant status badges.

j-bunk: Yep that makes sense, the only confusing part that led me to leave it out is that in the designs we can have an active badge with an icon and one without an icon (image below) and I wasn't sure if I should make 2 variants for that. So my thinking was anything with an icon would be in `ApplicantStatusBadge` otherwise it would use a normal badge. I could make another variant for an active badge without an icon and then have all the badges in `ApplicantStatusBadge` there if that makes it more consistent. Thoughts?

image

OustanDing: That's a really good point. In that case, let's just go with your idea - leave the badges without icons to be implemented using the Chakra UI `Badge` component.

Furthermore, the badges with icons are used for requests whereas those without icons are used for applicants. Could you please rename this component to `RequestStatusBadge` instead? Apologies for this being so late.

j-bunk: All good! Addressed!

OustanDing (approved):
LGTM! Thanks for the quick changes :)

PR #24, opened 2021-07-10
[Improvement] Update Database schemas

angeladietz:
## Notion ticket link

[Add Address Line 2 to all DB schemas](https://www.notion.so/uwblueprintexecs/Add-Address-Line-2-to-all-DB-schemas-683cc2f47a384c8482839c1ad504b0f0)
[Add gender text field to applicant, application for other genders](https://www.notion.so/uwblueprintexecs/Add-gender-text-field-to-applicant-application-for-other-genders-51057f8a097748cab828c2bca1f83643)
[Add acceptedTOC field to applicant model](https://www.notion.so/uwblueprintexecs/Add-acceptedTOC-field-to-applicant-model-15fe7832c40f418ba1cb6b60ec4e17d9)

## Implementation description
* Added new fields to schema.sql and schema.graphql files

## Notes
* Everyone will need to run `npx prisma db push` locally to update their local db schemas once this PR is merged
* Still a WIP because I haven't been able to get the schema.prisma file updated correctly. I think I may have a problem with my local db setup but if anyone has any tips on how to debug this so that the schema.prisma file will get updated please let me know!!

## Checklist
- [x] My PR name is descriptive, is in imperative tense and starts with one of the following: `[Feature]`,`[Improvement]` or `[Fix]`,
- [x] I have run the appropriate linter(s)
- [x] I have requested a review from the RCD team on GitHub, or specific people who are associated with this ticket

PR #25, opened 2021-07-11
Hookup GraphQL models in resolvers

jeffreyzhang2001:
## Notion ticket link

[Hook up GraphQL resolvers together for various models](https://www.notion.so/uwblueprintexecs/ac7b1f636a214a0cbf45f6b76f899b54?v=5414ee5fc5974f149921e192d6eb29b8&p=a4cbc48f95734f8b9a49f9ee01845504)

## Implementation description
* Added types for Guardian and MedicalInformation
* Added ability to query employees by ID (for future usage)
* Added eager loading on a bunch of queries with Prisma includes (e.g. fetch applicants' applications when querying for applicants)

## Notes
* Please look through the PR and let me know if there's anything I missed :D

## Checklist
- [x ] I have run the appropriate linter(s)
- [x ] I have requested a review from the RCD team on GitHub, or specific people who are associated with this ticket

### DIFF
--- a/lib/applicants/resolvers.ts
+++ b/lib/applicants/resolvers.ts
@@ -9,7 +9,12 @@ import { formatPhoneNumber, formatPostalCode } from '@lib/utils/format';
  * @returns All RCD applicants
  */
 export const applicants: Resolver = async (_parent, _args, { prisma }) => {
-  const applicants = await prisma.applicant.findMany();
+  const applicants = await prisma.applicant.findMany({
+    include: {
+      applications: true,
+      permits: true,
+    },
+  });
   return applicants;
 };
 
--- a/lib/applicants/resolvers.ts
+++ b/lib/applicants/resolvers.ts
@@ -13,6 +13,8 @@ export const applicants: Resolver = async (_parent, _args, { prisma }) => {
     include: {
       applications: true,
       permits: true,
+      guardian: true,
+      medicalInformation: true,
     },
   });
   return applicants;
--- a/lib/applicants/resolvers.ts
+++ b/lib/applicants/resolvers.ts
@@ -9,14 +9,7 @@ import { formatPhoneNumber, formatPostalCode } from '@lib/utils/format';
  * @returns All RCD applicants
  */
 export const applicants: Resolver = async (_parent, _args, { prisma }) => {
-  const applicants = await prisma.applicant.findMany({
-    include: {
-      applications: true,
-      permits: true,
-      guardian: true,
-      medicalInformation: true,
-    },
-  });
+  const applicants = await prisma.applicant.findMany();
   return applicants;
 };
 
--- a/lib/applicants/schema.graphql
+++ b/lib/applicants/schema.graphql
@@ -18,7 +18,7 @@ type Applicant {
   rcdUserId: Int
   acceptedTOC: Date
   # status: ApplicantStatus
-  # applications: [Application!]
+  applications: [Application!]
   # guardian: Guardian
   # medicalInformation: MedicalInformation
   # permits: [Permit!]
--- a/lib/applicants/schema.graphql
+++ b/lib/applicants/schema.graphql
@@ -17,11 +17,11 @@ type Applicant {
   postalCode: String!
   rcdUserId: Int
   acceptedTOC: Date
-  # status: ApplicantStatus
+  status: ApplicantStatus
   applications: [Application!]
-  # guardian: Guardian
-  # medicalInformation: MedicalInformation
-  # permits: [Permit!]
+  guardian: Guardian
+  medicalInformation: MedicalInformation
+  permits: [Permit!]
 }
 
 input CreateApplicantInput {
--- a/lib/applicants/schema.graphql
+++ b/lib/applicants/schema.graphql
@@ -1,5 +1,3 @@
-# TODO: Integrate commented out types (ApplicantStatus, Application, Guardian, MedicalInformation, Permit)
-
 type Applicant {
   id: ID!
   firstName: String!
@@ -40,11 +38,6 @@ input CreateApplicantInput {
   postalCode: String!
   rcdUserId: Int
   acceptedTOC: Date
-  # status: ApplicantStatus
-  # applications: [Application!]
-  # guardian: Guardian
-  # medicalInformation: MedicalInformation
-  # permits: [Permit!]
 }
 
 type CreateApplicantResult {
--- a/lib/applications/field-resolvers.ts
+++ b/lib/applications/field-resolvers.ts
@@ -1,6 +1,10 @@
 import { Resolver } from '@lib/resolvers'; // Resolver type
-import { Application } from '@lib/types';
+import { Application } from '@lib/types'; // Application type
 
+/**
+ * Field resolver to fetch the applicant that the application belongs to
+ * @returns Applicant object
+ */
 export const applicationApplicantResolver: Resolver<Application> = async (
   parent,
   _args,
@@ -9,6 +13,11 @@ export const applicationApplicantResolver: Resolver<Application> = async (
   return await prisma.applicant.findUnique({ where: { id: parent?.applicantId || undefined } });
 };
 
+/**
+ * Field resolver to fetch the permit associated with an application.
+ * NOTE/TODO: Each application should have exactly one permit. Fix schema.
+ * @returns Array of permit objects
+ */
 export const applicationPermitResolver: Resolver<Application> = async (
   parent,
   _args,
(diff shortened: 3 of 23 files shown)


lib/medicalInformation/schema.graphql
```diff
@@ -0,0 +1,12 @@
+type MedicalInformation {
```
angeladietz, line 1: Do we also need to be adding resolvers for the `MedicalInformation` and `Guardian`?

lib/applicants/schema.graphql
```diff
@@ -14,11 +14,11 @@ type Applicant {
   address: String!
   postalCode: String!
   rcdUserId: Int
-  # status: ApplicantStatus
```
angeladietz, line 17: You can also remove the TODO comment at the beginning of this file

lib/applications/resolvers.ts
```diff
@@ -28,7 +32,12 @@ export const createApplication: Resolver = async (_, args, { prisma }) => {
   let application;
   try {
     application = await prisma.application.create({
-      data: { ...args.input },
+      data: {
+        ...args.input,
+        applicant: {
+          connect: { id: applicantId },
```
angeladietz, line 38: Should we also be doing a `connect` like this in `lib/permits/resolvers`? 

Also, there's some TODOs in `lib/permits/schema.graphql` regarding the same task as this PR. I'm not totally sure if your changes already fix them or not but I think it would make sense for you to address them in this pr.

angeladietz (dismissed):
Looks great overall! Just left a few small comments.

lib/applicants/resolvers.ts
```diff
@@ -9,7 +9,14 @@ import { formatPhoneNumber, formatPostalCode } from '@lib/utils/format';
...
 export const applicants: Resolver = async (_parent, _args, { prisma }) => {
-  const applicants = await prisma.applicant.findMany();
+  const applicants = await prisma.applicant.findMany({
+    include: {
+      applications: true,
+      permits: true,
+      guardian: true,
+      medicalInformation: true,
+    },
+  });
```
OustanDing, line 19: We don't want to include all of these fields off the bat. The goal of GraphQL is to only do as much work as needed, so we don't want to do all these joins and queries if we're not asking for these fields.

Instead, we want to make use of field resolvers to retrieve these nested fields, like so (from Ben Awad video lol):
image

We'll address N+1 problem later.

lib/applications/resolvers.ts
```diff
@@ -12,7 +12,11 @@ import { DBErrorCode } from '@lib/db/errors'; // Database errors
  * @returns All RCD applications
  */
 export const applications: Resolver = async (_parent, _args, { prisma }) => {
-  const applications = await prisma.application.findMany();
+  const applications = await prisma.application.findMany({
+    include: {
+      permits: true,
+    },
+  });
```
OustanDing, line 19: Same here as above.

lib/applications/resolvers.ts
OustanDing: Yes we should have a `connect` when creating permits. We always want a created permit to be attached to an applicant.

lib/guardian/schema.graphql
```diff
@@ -0,0 +1,14 @@
+type Guardian {
+  id: ID!
+  firstName: String!
+  middleName: String
+  lastName: String!
+  address: String!
```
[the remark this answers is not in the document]
OustanDing: Good heads up. We probably just want to do change `address` to `addressLine1` and `addressLine2` in this PR then. I don't think it will raise conflicts.

lib/medicalInformation/schema.graphql
OustanDing: Not at the top level. However, we would want to have field resolvers for medical information and guardian so that they can be retrieved when querying for applicants. I mentioned this in my comment above.

lib/physicians/resolvers.ts
```diff
@@ -10,7 +10,11 @@ import { formatPhoneNumber, formatPostalCode } from '@lib/utils/format';
  * @returns All physicians of RCD applicants
  */
 export const physicians: Resolver = async (_parent, _args, { prisma }) => {
-  const physicians = await prisma.physician.findMany();
+  const physicians = await prisma.physician.findMany({
+    include: {
+      medicalInformation: true,
+    },
+  });
```
OustanDing, line 17: Same comment as above.

OustanDing (changes requested):

lib/guardian/schema.graphql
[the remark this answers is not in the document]
OustanDing: Actually, I think we can still merge #24 first, then rebase. This PR doesn't have any database changes, and the GraphQL schema changes depend on the SQL schemas.

jeffreyzhang2001:
Addressed all the comments, I know it's a bit of a big PR but there are a bunch of patterns across the changes so it shouldn't be too bad.

Note: You can re-seed your database with the seed script included in this PR (`npx prisma db seed --preview-feature`)

lib/applications/field-resolvers.ts
```diff
@@ -0,0 +1,18 @@
...
+) => {
+  return await prisma.applicant.findUnique({ where: { id: parent?.applicantId || undefined } });
+};
+
+export const applicationPermitResolver: Resolver<Application> = async (
+  parent,
+  _args,
+  { prisma }
+) => {
+  return await prisma.permit.findMany({ where: { applicationId: parent?.id } });
```
OustanDing, line 17: This should be `findUnique` since every application can only be related to at most one permit.

lib/employees/resolvers.ts
```diff
@@ -4,8 +4,8 @@ import { EmployeeAlreadyExistsError } from '@lib/employees/errors'; // Employee
 import { DBErrorCode } from '@lib/db/errors'; // Database errors
 
 /**
- * Query all the RCD employees in the internal-facing app
- * @returns All RCD employees
+ * Queries for RCD employees in the internal-facing app
+ * @returns An array of RCD employees
```
OustanDing, line 8: We can revert this comment.

lib/guardian/schema.graphql
```diff
@@ -0,0 +1,15 @@
+type Guardian {
+  id: ID!
+  firstName: String!
+  middleName: String
+  lastName: String!
+  addressLine1: String!
+  addressLine2: String!
```
OustanDing, line 7:
```suggestion
  addressLine2: String
```

lib/guardian/schema.graphql
```diff
@@ -0,0 +1,15 @@
...
+  city: String!
+  province: Province!
+  postalCode: String!
+  phone: String!
+  relationship: String!
+  notes: String
+  applicant: Applicant!
```
OustanDing, line 14: I don't think this should be included in the schema, since we don't have a field resolver for `Applicant` under `Guardian`. Also, we wouldn't be querying for that anyway.

lib/permits/resolvers.ts
```diff
@@ -21,30 +21,42 @@ export const permits: Resolver = async (_parent, _args, { prisma }) => {
...
-      data: { ...input },
+      data: {
+        ...args.input,
+        applicant: {
+          connect: { id: applicantId },
+        },
+        application: {
+          connect: { id: applicationId },
+        },
+      },
```
OustanDing, line 39: Nice :D

prisma/seed.ts
```diff
@@ -1,14 +1,18 @@
 import prisma from '../prisma/index'; // Relative path required, path aliases throw error with seed command
-import { Role } from '@prisma/client';
-
+import { Role, Gender, Province, PaymentType, Aid } from '../lib/graphql/types';
+/**
+ * Main program
+ */
```
OustanDing, line 5:
```suggestion
import { Role, Gender, Province, PaymentType, Aid } from '../lib/graphql/types'; // Relative path required

/**
 * Main program
 */
```

OustanDing (changes requested):
Mostly documentation-related comments. The field resolvers look great and make sense! After resolving the issues, we should be good to go.

lib/applications/field-resolvers.ts
jeffreyzhang2001: The way our schema is defined, the `Application` model has multiple `Permit` objects associated with it. To fix this, we would have to modify our schema and change a bunch of types. For sake of getting an MVP out, I think it'll be fine to leave as is and fix later - `findMany` with the ID restriction will still work as desired, since each `Permit` is only associated with one object.

OustanDing: Good point. Nice catch! I will add a ticket to the backlog to address this post-MVP.

OustanDing (approved):
LGTM! Thank you for addressing the PR comments quickly!

PR #34, opened 2021-07-25
[Feature] Build Requests Table Page

angeladietz:
## Notion ticket link

[Request Table page](https://www.notion.so/uwblueprintexecs/Requests-Table-page-e6290e4f165e464283ec249fb8b062f9)

## Implementation description
* Front end for Request Table Page

Screen Shot 2021-07-24 at 11 01 51 PM

## Notes
* populating table with real data and applying filters will be done as part of the related API hookup ticket

## Checklist

### DIFF
--- a/pages/admin/index.tsx
+++ b/pages/admin/index.tsx
@@ -28,7 +28,14 @@ import Pagination from '@components/internal/Pagination'; // Pagination componen
 import { RequestStatusBadge } from '@components/internal/RequestStatusBadge'; //Status badge component
 
 type StatusProps = {
-  value: any;
+  readonly value:
+    | 'completed'
+    | 'inProgress'
+    | 'pending'
+    | 'rejected'
+    | 'expiring'
+    | 'expired'
+    | 'active';
 };
 
 function renderStatusBadge({ value }: StatusProps) {
@@ -40,7 +47,10 @@ function renderStatusBadge({ value }: StatusProps) {
 }
 
 type NameProps = {
-  readonly value: any;
+  readonly value: {
+    name: string;
+    id: string;
+  };
 };
 
 function renderName({ value }: NameProps) {
@@ -138,7 +148,7 @@ export default function Requests() {
       <GridItem colSpan={12}>
         <Flex justifyContent="space-between" alignItems="center" marginBottom="32px">
           <Text textStyle="display-xlarge">Requests</Text>
-          {/* TODO
+          {/* TODO: 'Create a new request' function is out of scope for this term's MVP
           <Menu>
             <MenuButton
               as={Button}
--- a/pages/admin/index.tsx
+++ b/pages/admin/index.tsx
@@ -16,6 +16,7 @@ import {
   InputGroup,
   Input,
   InputLeftElement,
+  useControllableState,
 } from '@chakra-ui/react'; // Chakra UI
 import { AddIcon, ChevronDownIcon, SearchIcon } from '@chakra-ui/icons'; // Chakra UI Icons
 import Layout from '@components/internal/Layout'; // Layout component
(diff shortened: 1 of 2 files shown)


pages/admin/index.tsx
```diff
@@ -11,80 +11,149 @@ import {
...
     requestType: 'Replacement',
-    status: 'In progress',
+    status: 'completed',
   },
 ];
 
 // Internal home page - view APP requests
 export default function Requests() {
+  const [permitTypeFilter, setPermitTypeFilter] = useControllableState({ defaultValue: 'All' });
+  const [requestTypeFilter, setRequestTypeFilter] = useControllableState({ defaultValue: 'All' });
```
EmilioMena, line 134: just curious why did you opt to the `useControllableState` hook as opposed to `useState`?

pages/admin/index.tsx
```diff
@@ -11,80 +11,149 @@ import {
...
+
   return (
     <Layout>
       <GridItem colSpan={12}>
         <Flex justifyContent="space-between" alignItems="center" marginBottom="32px">
           <Text textStyle="display-xlarge">Requests</Text>
-          <Button leftIcon={<AddIcon />} height="48px" padding="24px">
-            Create a request
-          </Button>
+          {/* TODO
```
EmilioMena, line 141: can you add something about the create function being addressed post mvp. It took me a minute to remember why this was commented out, so i can only imagine what its going too be like looking at this again a couple of months 😅

pages/admin/index.tsx
```diff
@@ -11,80 +11,149 @@ import {
...
 import Layout from '@components/internal/Layout'; // Layout component
 import { Role } from '@lib/types'; // Role enum
 import { authorize } from '@tools/authorization'; // Page authorization
 import Table from '@components/internal/Table'; // Table component
 import Pagination from '@components/internal/Pagination'; // Pagination component
+import { RequestStatusBadge } from '@components/internal/RequestStatusBadge'; //Status badge component
+
+type StatusProps = {
+  value: any;
+};
```
EmilioMena, line 32: since status can only be a discrete set of values (variants in RequestStatusBadge) I think it would be a good idea to add those as a type instead of using `any`. Also it should be `readonly`

pages/admin/index.tsx
```diff
@@ -11,80 +11,149 @@ import {
...
+  return (
+    <Wrap>
+      <RequestStatusBadge variant={value}></RequestStatusBadge>
+    </Wrap>
+  );
+}
+
+type NameProps = {
+  readonly value: any;
+};
```
EmilioMena, line 44: similar thing here instead of `any`

```
value: {
          name: string;
           id: number;
       };
```

pages/admin/index.tsx
```diff
@@ -11,80 +11,149 @@ import {
...
 import Pagination from '@components/internal/Pagination'; // Pagination component
+import { RequestStatusBadge } from '@components/internal/RequestStatusBadge'; //Status badge component
+
+type StatusProps = {
+  value: any;
+};
+
+function renderStatusBadge({ value }: StatusProps) {
+  return (
+    <Wrap>
```
EmilioMena, line 36: hm correct me if i'm wrong but i don't think you need this `<Wrap>` here

EmilioMena (commented):
Good job y'all! Left a few small comments but other than that ready to 🚢

pages/admin/index.tsx
EmilioMena: nvm i see what happens without it

jennifertsai:
Just addressed your comments @EmilioMena !! Angela is going to reply to your question about `useControllableState` soon as well :)

EmilioMena (approved):

PR #37, opened 2021-07-26
[Feature] Create permit holders table page

EmilioMena:
## Notion ticket link

[Permit Holders Table Page](https://www.notion.so/uwblueprintexecs/Permit-Holders-Table-Page-52ce53d70edb417cbdb3352f0ae265ce)

## Implementation description
- create table in permit holders page
- filter, search and sorting
- react-day-picker for calendar dropdown and react-helmet for styling calendar

image
image

## Notes
- added react helmet based on the[ date picker docs](http://react-day-picker.js.org/examples/selected-range#:~:text=%3CHelmet%3E%0A%20%20%20%20%20%20%20%20%20%20%3Cstyle%3E%7B%60%0A%20%20.Selectable,important%3B%0A%20%20%7D%0A%60%7D%3C/style%3E%0A%20%20%20%20%20%20%20%20%3C/Helmet%3E)

## Checklist

### DIFF
--- a/pages/admin/permit-holders.tsx
+++ b/pages/admin/permit-holders.tsx
@@ -1,16 +1,260 @@
 import { GetServerSideProps } from 'next'; // Get server side props
 import { getSession } from 'next-auth/client'; // Session management
-import { Text, GridItem } from '@chakra-ui/react'; // Chakra UI
+import {
+  Box,
+  Flex,
+  Text,
+  GridItem,
+  Button,
+  Menu,
+  MenuButton,
+  InputGroup,
+  Input,
+  InputLeftElement,
+  useControllableState,
+  MenuList,
+  MenuItem,
+  Wrap,
+  Badge,
+} from '@chakra-ui/react'; // Chakra UI
+import { ChevronDownIcon, SearchIcon } from '@chakra-ui/icons'; // Chakra UI Icons
 import Layout from '@components/internal/Layout'; // Layout component
 import { Role } from '@lib/types'; // Role enum
 import { authorize } from '@tools/authorization'; // Page authorization
+import Table from '@components/internal/Table'; // Table component
+import Pagination from '@components/internal/Pagination'; // Pagination component
+import { SetStateAction, Dispatch } from 'react';
+import DayPicker, { DateUtils, DayPickerProps, RangeModifier } from 'react-day-picker';
+import 'react-day-picker/lib/style.css';
+
+// Placeholder data
+
+const DATA = Array(10).fill({
+  name: {
+    name: 'Charmaine Wang',
+    id: 36565,
+  },
+  dateOfBirth: '2021/01/01',
+  homeAddress: {
+    address: '5300, No.3 Rd Lansdowne Centre',
+    city: 'Richmond',
+    postalCode: 'V2X 1E4',
+  },
+  email: 'charmainewang@rcd.org',
+  phone: '000-000-0000',
+  recentAPP: '#XXXXXX',
+  userStatus: 'active',
+});
+
+const COLUMNS = [
+  {
+    Header: 'Name',
+    accessor: 'name',
+    Cell: renderName,
+  },
+  {
+    Header: 'Date of Birth',
+    accessor: 'dateOfBirth',
+    disableSortBy: true,
+  },
+  {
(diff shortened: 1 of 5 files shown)


pages/admin/permit-holders.tsx
```diff
@@ -1,16 +1,344 @@
...
+    id: number;
+  };
+};
+
+function renderName({ value }: NameProps) {
+  return (
+    <>
+      <Text>{value.name}</Text>
+      <Text textStyle="caption" textColor="secondary">
+        {value.id}
```
jennifertsai, line 105: Just a small fix here according to the designs: should be `ID: {value.id}` instead of just `{value.id}`

pages/admin/permit-holders.tsx
```diff
@@ -1,16 +1,344 @@
...
+                <MenuButton
+                  as={Button}
+                  variant="outline"
+                  rightIcon={<ChevronDownIcon />}
+                  marginRight="12px"
+                  color="text.secondary"
+                  borderColor="border.secondary"
+                >
+                  <MenuText name={`User Status`} value={userStatusFilter} />
+                </MenuButton>
```
jennifertsai, line 300: Need to confirm this with designers but could we make the menu buttons a fixed size? Currently, the width of the menu box changes according to the options in the dropdown menu selected.

For example, this is what we see in Figma (there is more space after the "all")
image

Compared to what we have right now (no space after "all" and button size is fit to text length):
image

pages/admin/permit-holders.tsx
```diff
@@ -1,16 +1,344 @@
...
+  email: 'charmainewang@rcd.org',
+  phone: '000-000-0000',
+  recentAPP: {
+    appNumber: '#XXXXXX',
+    expiryDate: new Date(),
+  },
+  userStatus: 'active',
+});
+
+const COLUMNS = [
```
jennifertsai, line 52: The widths of the columns should be set according to the Figma specs:
image

It's pretty simple, you can check out Angela and I's PR, we have an example of how to do that: https://github.com/uwblueprint/richmond-centre-for-disability/pull/34/files

jennifertsai (commented):
This looks awesome!! I've added a few small comments just related to the design of the components :)

Just wanted to ask why we left out the "Actions" column? Is that currently out of scope for this term?

pages/admin/permit-holders.tsx
```diff
@@ -1,16 +1,344 @@
...
+
+  const dateRangeString = () => {
+    if (!range.to) {
+      return 'YYYY-MM-DD - YYYY-MM-DD';
+    }
+    if (range.from && range.to) {
+      return `${range.from.toLocaleDateString('en-CA')} - ${range.to.toLocaleDateString('en-CA')}`;
+    }
+  };
+
```
jennifertsai, line 262: Not sure if we could do this with react date picker, but when I was playing around with the date picker, I found it a bit unintuitive that I couldn't easily change the first date picked. For example, if a user selected this at first:
image

But then realized they made a mistake and meant to select July 19 instead of July 17 (first date in the range), the date picker I think defaults to modifying the second date when I click on July 19:
image

Is there a way to specify which date to change?

jennifertsai: Also, we could maybe consider adding some logic that won't allow the user to select a date prior to present day. Don't think that's super important though

pages/admin/permit-holders.tsx
EmilioMena: Good catch! Made it fixed width

pages/admin/permit-holders.tsx
EmilioMena: Both great points, to keep this PR short i think it's best to make another ticket for that and it can be addressed after the api hookup or post-mvp. For posterity here is a good example on the react-day-picker docs that we can follow to do your first suggestion http://react-day-picker.js.org/examples/selected-range-enter

EmilioMena: cc @OustanDing just so you're in the loop

EmilioMena:
Thanks for the comments @jennifertsai 🚀 Ready for another round!

jennifertsai (approved):
LGTM!! 🚀

PR #40, opened 2021-07-28
[Improvement] Improved generate-graphql-types script

OustanDing:
## Notion ticket link

[Ticket](https://www.notion.so/uwblueprintexecs/Improve-generate-graphql-types-script-41cef4a02c644a56a60c0c3560b94e15)

## Implementation description
* Output errors to `stdout`
* Create a backup `types-backup.ts` file before deleting `types.ts`. If execution fails, restore the `types.ts` file using the backup file. This eliminates the cycle that arose from deleting `types.ts` unconditionally, which disabled the app from running, which in turn meant that we could not generate GraphQL types.
* Added `lib/graphql/types.ts` to `.prettierignore` so that we don't accidentally make changes to it

## Notes
* QOL change :))

## Checklist


lib/scripts/generate-graphql-types.js
```diff
@@ -8,18 +8,40 @@ const { execSync } = require('child_process'); // Execute shell commands
...
+    execSync('graphql-codegen --config codegen.yml', { stdio: 'inherit' });
+    console.log(`TypeScript types file created at ${TYPES_PATH} ✅\n`);
+
+    // Delete backup if it was made
+    if (existsSync(BACKUP_TYPES_PATH)) {
+      unlinkSync(BACKUP_TYPES_PATH);
+    }
 
-  console.log('DONE ✅');
+    console.log('DONE ✔️');
```
angeladietz, line 35: Just wondering why you changed checkmarks? Personally I liked the green checkmark better since its a lot easier to see in a black terminal but it doesn't really matter 😅

lib/graphql/types.ts
```diff
@@ -319,26 +320,32 @@ export type Mutation = {
   createPermit: CreatePermitResult;
 };
 
+
```
angeladietz, line 323: It would probably be nice to not add all of this extra whitespace (unless it was intentional?)

lib/scripts/generate-graphql-types.js
OustanDing: Just wanted to make it consistent with the red X for failure - will change it back though, kinda liked the original one more too xD

lib/graphql/types.ts
OustanDing: These whitespaces were automatically generated by the script! I reran it so that any manual changes to `types.ts` get overwritten.

jeffreyzhang2001 (approved):

PR #42, opened 2021-08-02
[Feature] Permit Holder Table API Hookup

EmilioMena:
## Notion ticket link

[Permit Holder page API hookup (has table)](https://www.notion.so/uwblueprintexecs/ac7b1f636a214a0cbf45f6b76f899b54?v=5414ee5fc5974f149921e192d6eb29b8)

## Implementation description
- [x] Add search, filter to applicants resolver
- [x] Add recentPermit field to applicants resolver (fetch the most recent permit)
- [x] Create frontend graphql queries
- [x] Update permit-holders table with required event handlers

## Notes

I've added some seed data to make it easier to test locally. If you're using heroku you can reset your db by running
- reset db => `heroku pg:reset --app <your-app-name>`
- create db => `heroku pg:psql -a <your-app-name>  -f prisma/schema.sql`
- seed db => `npx prisma db seed --preview-feature`

## Checklist


lib/applicants/resolvers.ts
```diff
@@ -8,14 +8,146 @@ import {
...
+    skip: skip,
+    take: take,
+    orderBy: [
+      { firstName: sortingOrder.name || SortOrder.ASC },
+      { lastName: sortingOrder.name || SortOrder.ASC },
+    ],
+  });
+
+  return {
+    node: applicants,
```
angeladietz, line 148: Maybe we could rename `node` to something a bit more descriptive?

lib/graphql/resolvers.ts
```diff
@@ -15,7 +15,8 @@ import {
   applicantGuardianResolver,
   applicantMedicalInformationResolver,
   applicantMedicalHistoryResolver,
-} from '@lib/applicants/field-resolvers'; // Applicant field resolvers
+  applicantRecentPermitResolver,
+} from '@lib/applicants/field-resolvers';
```
angeladietz, line 19: Bring back the import comment :)

pages/_app.tsx
```diff
@@ -12,6 +12,7 @@ import '@fontsource/noto-sans/700.css'; // Noto sans bold
 const apolloClient = new ApolloClient({
   uri: '/api/graphql',
   cache: new InMemoryCache(),
+  connectToDevTools: true,
```
angeladietz, line 15: Just wondering, what does this do?

pages/admin/permit-holders.tsx
```diff
@@ -18,43 +19,29 @@ import {
...
+import {
+  FILTER_PERMIT_HOLDERS_QUERY,
+  FilterPermitHoldersRequest,
+  FilterPermitHoldersResponse,
+} from '@tools/pages/permit-holders/filter-permit-holders';
 import DayPicker, { DateUtils, DayPickerProps, RangeModifier } from 'react-day-picker'; // Date picker
 import 'react-day-picker/lib/style.css'; // Date picker styling
 import Helmet from 'react-helmet'; // Date picker inline styling for select range functionality
+import { SortOptions, SortOrder } from '@tools/types';
+import { Column } from 'react-table';
```
angeladietz, line 36: Comment imports in this file

pages/admin/permit-holders.tsx
```diff
@@ -88,35 +78,42 @@ const COLUMNS = [
...
-    accessor: 'userStatus',
+    accessor: 'status',
     disableSortBy: true,
     width: 120,
     maxWidth: 120,
     Cell: renderUserStatusBadge,
   },
 ];
 
+const PAGE_SIZE = 1;
```
angeladietz, line 97: Same thing as me, change this to 20 before merging

pages/admin/permit-holders.tsx
```diff
@@ -141,64 +139,55 @@ function renderAddress({ value }: AddressProps) {
...
+  const expired = DateUtils.isPastDay(expiryDate);
+  const expiresSoon = DateUtils.isDayInRange(expiryDate, {
     from: today,
     to: DateUtils.addMonths(today, 1),
   });
   return (
     <Flex>
       <Text as="span" mr="9px">
-        {value.appNumber}
+        # {value.rcdPermitId}
```
angeladietz, line 171: Remove the space after the #

pages/admin/permit-holders.tsx
```diff
@@ -294,13 +343,23 @@ export default function PermitHolders() {
...
-                  <DropDownItems
-                    items={permitStatusOptions}
-                    setStateCallback={setPermitStatusFilter}
-                  />
+                  <MenuItem onClick={() => setPermitStatusFilter(undefined)}>All</MenuItem>
+                  {permitStatusOptions.map((value, i) => (
+                    <MenuItem
+                      key={`dropDownItem-${i}`}
+                      onClick={() => {
+                        setPermitStatusFilter(value);
```
angeladietz, line 357: I think that we need to also reset the page number to 0 anytime a filter is applied

lib/applicants/resolvers.ts
```diff
@@ -8,14 +8,146 @@ import {
...
+                  lte: expiryDateRangeTo?.toISOString(),
+                },
+              },
+            ],
+          },
+        }
+      : undefined;
+
+    // Update default filter since there were filter arguments
+    where = {
```
angeladietz, line 114: I think that another filter is required here to only return applicants that have a "recent permit", as in they have at least one permit associated with them. Not 100% this is even a valid scenario, but it does seem reasonable that an applicant could be created before a permit is associated with them. I got some errors in `pages/admin/permit-holders` due to getting null results for the `recentPermit` and `status` fields.

angeladietz (changes requested):
Looks awesome overall! Just left a couple small notes

pages/admin/permit-holders.tsx
```diff
@@ -141,64 +139,55 @@ function renderAddress({ value }: AddressProps) {
...
+type UserStatusBadgeProps = {
   value: 'active' | 'inactive';
 };
 
-function renderUserStatusBadge({ value }: UserStatusProps) {
+// User status cell in table
+function renderUserStatusBadge({ value }: UserStatusBadgeProps) {
   return (
     <Wrap>
       <Badge variant={value}>{value.toUpperCase()}</Badge>
```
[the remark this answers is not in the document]
EmilioMena: good catch! hm i wonder why an applicant's status can be null. Doesn't seem like it should be

pages/_app.tsx
EmilioMena: 😅 i added apollo dev tools while debugging which needed this line and forgot to remove it, nice catch!

lib/applicants/resolvers.ts
EmilioMena: Good point! I double checked with Oustan abt this and it turns out every applicant entry must have at least one associated permit since to become an 'applicant' they'd need to have their application approved which gives them a permit in the process. I changed the seed data to better reflect this

As a sanity check tho I also added a clause to make sure every applicant has at least one permit and changed the gql schema too in my latest commit

pages/admin/permit-holders.tsx
EmilioMena: yeah i missed that good catch

EmilioMena: When trying to implement this in the way you did it on the request table page I'm getting some weird results so I'm pairing with oustan later to figure it out.
It'll be added before merging

components/internal/Table.tsx
```diff
@@ -37,7 +37,7 @@ export default function Table(props: Props) {
     if (onChangeSortOrder !== undefined) {
       onChangeSortOrder(getSortOptions(sortBy));
     }
-  }, [onChangeSortOrder, sortBy]);
+  }, [sortBy]);
```
OustanDing, line 40: Any reason why `onChangeSortOrder` was removed here? I think our linting should catch that :/

components/internal/Table.tsx
```diff
@@ -82,13 +82,13 @@ export default function Table(props: Props) {
...
             return (
               <Tr {...row.getRowProps()} key={row.id}>
-                {row.cells.map(cell => (
+                {row.cells.map((cell, i) => (
                   <Td
                     height="80px"
                     paddingX="0"
                     color="text.secondary"
                     {...cell.getCellProps()}
-                    key={row.id}
```
OustanDing, line 91: What was the reason for this change? I feel like the key might get duplicated because we're calling this `map` function on every row, which means the `i` values will repeat.

lib/applicants/field-resolvers.ts
```diff
@@ -18,7 +19,11 @@ export const applicantApplicationsResolver: Resolver<Applicant> = async (
  * @returns Array of permit objects
  */
 export const applicantPermitsResolver: Resolver<Applicant> = async (parent, _args, { prisma }) => {
-  return await prisma.permit.findMany({ where: { applicantId: parent?.id } });
+  return await prisma.applicant
+    .findUnique({
+      where: { id: parent.id },
+    })
+    .permits();
```
OustanDing, line 26: Does this work better in general? Just wondering if there are any tradeoffs between the two approaches.

lib/applicants/resolvers.ts
```diff
@@ -8,14 +8,142 @@ import {
...
+ * - expiryDateRangeTo: Permit Expiry Date end
+ * - search: Search by first, middle, last name or by RCD user ID
+ *
+ * Pagination:
+ * - offset: Number of results to skip.
+ * - limit: Number of result to return
+ *
+ * Sorting:
+ * - order: array of tuples of the field being sorted and the order. Default [['firstName', 'asc'], ['lastName', 'asc']]
+ * @returns All RCD applicants that match the filter(s).
```
OustanDing, line 32: Great documentation :D

lib/applicants/resolvers.ts
```diff
@@ -8,14 +8,142 @@ import {
...
+
+  if (filter?.order) {
+    filter.order.forEach(([field, order]: [string, SortOrder]) => (sortingOrder[field] = order));
+  }
+
+  const take = filter?.limit;
+  const skip = filter?.offset;
+
+  const totalCount = await prisma.applicant.count({
+    where: where,
```
OustanDing, line 130:
```suggestion
    where,
```
and same below

lib/applicants/schema.graphql
```diff
@@ -75,3 +76,19 @@ input UpdateApplicantInput {
 type UpdateApplicantResult {
   ok: Boolean!
 }
+
+input ApplicantsFilter {
+  order: [[String]]
```
OustanDing, line 81: Just want to double check, is there a way to specify that the inner array has 2 elements of type `String` and `SortOrder`?

lib/applicants/schema.graphql
```diff
@@ -75,3 +76,19 @@ input UpdateApplicantInput {
...
+  userStatus: UserStatus
+  expiryDateRangeFrom: Date
+  expiryDateRangeTo: Date
+  search: String
+  limit: Int
+  offset: Int
+}
+
+type QueryApplicantsResult {
+  result: [Applicant]!
```
OustanDing, line 92:
```suggestion
  result: [Applicant!]!
```
since every `Applicant` must be non-null

lib/graphql/resolvers.ts
```diff
@@ -57,6 +58,7 @@ const resolvers = {
     guardian: applicantGuardianResolver,
     medicalInformation: applicantMedicalInformationResolver,
     medicalHistory: applicantMedicalHistoryResolver,
+    recentPermit: applicantRecentPermitResolver,
```
OustanDing, line 61: Sorry this is so far down in the review lol but can we rename this to `mostRecentPermit`? Makes it more clear

lib/graphql/schema.graphql
```diff
@@ -1,6 +1,6 @@
 type Query {
   meta: Meta!
-  applicants: [Applicant!]
+  applicants(filter: ApplicantsFilter): QueryApplicantsResult!
```
OustanDing, line 3:
```suggestion
  applicants(filter: ApplicantsFilter): QueryApplicantsResult
```
If user is not authorized, it will return `null`

lib/graphql/schema.graphql
```diff
@@ -90,3 +90,14 @@ enum ReasonForReplacement {
   STOLEN
   OTHER
 }
+
+enum PermitStatus {
+  VALID
+  EXPIRED
+  EXPIRING_THIRTY
```
OustanDing, line 97:
```suggestion
  EXPIRING_IN_THIRTY_DAYS
```
Better to be more clear in variable names to reduce potential confusion

pages/admin/index.tsx
```diff
@@ -272,7 +272,12 @@ export default function Requests() {
             </Flex>
             <Table columns={COLUMNS} data={DATA} />
             <Flex justifyContent="flex-end">
-              <Pagination pageSize={20} totalCount={150} />
+              <Pagination /* eslint-disable @typescript-eslint/no-empty-function */
+                pageNumber={0}
+                onPageChange={() => {}}
```
OustanDing, line 277: cc @angeladietz @jennifertsai we moved the state management for the page number to be outside of the `Pagination` component, so you will need to define this state yourself and pass it into `Pagination`.

prisma/dev-seed-utils/permits.ts
```diff
@@ -3,19 +3,40 @@
...
 // Seed data
-const permits = [{ rcdPermitId: 1, applicantId: 1, applicationId: 1 }];
+const permits = [
+  {
+    rcdPermitId: 1,
+    applicantId: 1,
+    applicationId: 1,
+    expiryDate: new Date(2025, 1, 1).toISOString(),
+    active: true,
+  },
```
OustanDing, line 13: Thanks @EmilioMena for taking the time to add extra seeding data! Will help us a lot for testing

prisma/index.ts
```diff
@@ -8,7 +8,11 @@ interface CustomNodeJsGlobal extends NodeJS.Global {
 // Prevent multiple instances of Prisma Client in development
 declare const global: CustomNodeJsGlobal;
 
-const prisma = global.prisma || new PrismaClient();
+const prisma =
+  global.prisma ||
+  new PrismaClient({
+    log: ['query', 'info', `warn`, `error`],
```
OustanDing, line 14: Make sure to remove this

tools/components/internal/table.ts
```diff
@@ -3,7 +3,7 @@ import { SortOrder } from '@tools/types'; // Sort order enum
 
 // Function type of getSortOptions - Disable eslint preventing `object` from being used
 // eslint-disable-next-line
-type GetSortOptions = (sortBy: SortingRule<object>[]) => ReadonlyArray<[string, SortOrder]>;
+type GetSortOptions = (sortBy: SortingRule<object>[]) => Array<[string, SortOrder]>;
```
OustanDing, line 6: Just wondering, why did we change this to `Array`?

tools/hooks/useDebounce.ts
```diff
@@ -0,0 +1,26 @@
+import { useState, useEffect } from 'react'; // React
+
+// From usehooks.com/useDebounce
+export default function useDebounce<T>(value: T, delay: number) {
```
OustanDing, line 4: Sorry @EmilioMena looks like we forgot to add the return type for the hook. Could you please add it in a similar fashion to how it's done in our other hooks?

tools/pages/permit-holders/filter-permit-holders.ts
```diff
@@ -0,0 +1,38 @@
+import { gql } from '@apollo/client'; // gql tag
+import { ApplicantsFilter, Applicant } from '@lib/graphql/types';
+
+export const FILTER_PERMIT_HOLDERS_QUERY = gql`
```
OustanDing, line 4:
```suggestion
export const GET_PERMIT_HOLDERS_QUERY = gql`
```
Still a get request at the end of the day

tools/pages/permit-holders/filter-permit-holders.ts
```diff
@@ -0,0 +1,38 @@
...
+  }
+`;
+
+export type FilterPermitHoldersRequest = {
+  filter: ApplicantsFilter;
+};
+
+export type FilterPermitHoldersResponse = {
+  applicants: {
+    result: [Applicant];
```
OustanDing, line 35: I think this will be an array of one single `Applicant` lol

We want to use TypeScript's `Pick` so that we can get type checking and make sure that we are exactly receiving the fields we want. So it would be something like:

```suggestion
    result: ReadonlyArray<
      Pick<
        Applicant,
        | 'firstName'
        | 'middleName'
        | 'lastName'
        ...
      >
    >;
```

tools/types.ts
```diff
@@ -8,4 +8,4 @@ export enum SortOrder {
 }
 
 // Sort options type that contains an array of tuples of the field being sorted and the order
-export type SortOptions = ReadonlyArray<[string, SortOrder]>;
+export type SortOptions = Array<[string, SortOrder]>;
```
OustanDing, line 11: Same here, wondering why this was changed to `Array`?

pages/admin/permit-holders.tsx
```diff
@@ -18,43 +19,31 @@ import {
...
-});
-
-const COLUMNS = [
+const COLUMNS: Column<any>[] = [
   {
     Header: 'Name',
     accessor: 'name',
     width: 180,
     minWidth: 180,
+    sortDescFirst: true,
```
OustanDing, line 46: Isn't this supposed to be `false`?

pages/admin/permit-holders.tsx
```diff
@@ -63,6 +52,9 @@ const COLUMNS = [
     disableSortBy: true,
     width: 140,
     maxWidth: 140,
+    Cell: ({ value }: any) => {
+      return <Text>{new Date(value).toLocaleDateString('en-ZA')}</Text>;
```
OustanDing, line 56: What is `en-ZA`?

pages/admin/permit-holders.tsx
```diff
@@ -88,35 +80,42 @@ const COLUMNS = [
...
-    id: number;
+    firstName: string;
+    middleName: string;
+    lastName: string;
+    rcdUserId: number;
   };
 };
 
+// Name cell in table
 function renderName({ value }: NameProps) {
```
OustanDing, line 111: Can just put this in the cells configuration above

pages/admin/permit-holders.tsx
```diff
@@ -88,35 +80,42 @@ const COLUMNS = [
...
 ];
 
+const PAGE_SIZE = 20;
+
 type NameProps = {
   value: {
-    name: string;
-    id: number;
+    firstName: string;
+    middleName: string;
```
OustanDing, line 104: `middleName` could be undefined.

pages/admin/permit-holders.tsx
```diff
@@ -88,35 +80,42 @@ const COLUMNS = [
...
   };
 };
 
+// Name cell in table
 function renderName({ value }: NameProps) {
   return (
     <>
-      <Text>{value.name}</Text>
+      <Text>
+        {value.firstName} {value.middleName} {value.lastName}
```
OustanDing, line 115: Careful here. `middleName` could be `undefined` and with the way JS works, it will literally render `'undefined'` lol

So we should have something like
```suggestion
        {value.firstName} {value.middleName || ''} {value.lastName}
```

pages/admin/permit-holders.tsx
```diff
@@ -141,64 +141,55 @@ function renderAddress({ value }: AddressProps) {
...
   value: 'active' | 'inactive';
 };
 
-function renderUserStatusBadge({ value }: UserStatusProps) {
+// User status cell in table
+function renderUserStatusBadge({ value }: UserStatusBadgeProps) {
   return (
     <Wrap>
-      <Badge variant={value}>{value.toUpperCase()}</Badge>
+      <Badge variant={value}>{value?.toUpperCase()}</Badge>
```
OustanDing, line 152: Same thing here with the `undefined` issue.

Also, can just throw this render function in the cells configuration.

pages/admin/permit-holders.tsx
```diff
@@ -349,13 +449,31 @@ export default function PermitHolders() {
...
+                    placeholder="Search by first name, last name or user ID"
+                    onChange={event => {
+                      setSearchFilter(event.target.value);
+                    }}
+                  />
                 </InputGroup>
               </Box>
             </Flex>
-            <Table columns={COLUMNS} data={DATA} />
+            {!loading && networkStatus !== NetworkStatus.refetch && (
```
OustanDing, line 461: Can we just remove this condition for now? Always render the `Table`

OustanDing (changes requested):
Some changes and preemptive bug fixes. Really appreciate all the great work @EmilioMena !

lib/applicants/field-resolvers.ts
EmilioMena: yeah it's a prisma workaround to avoid n+1 queries. For some reason prisma doesn't batch findMany queries but it batches findUnique so this is the easiest way to not make a separate db query for every permit when we can just make 1 

https://www.prisma.io/docs/guides/performance-and-optimization/query-optimization-performance#solving-the-n1-problem

lib/applicants/schema.graphql
EmilioMena: true i guess there should always be applicants in the db

pages/admin/permit-holders.tsx
EmilioMena: `"yyyy/MM/dd"`

https://stackoverflow.com/questions/2388115/get-locale-short-date-format-using-javascript#:~:text=%22en-za%22%3A%20%22yyyy%2Fmm%2Fdd%22%2C

components/internal/Table.tsx
EmilioMena: It was making the page super super super slow 🐢 this is the error i saw in the console
image

I was act wondering why we had the sort order handler function in the dependencies to begin with?

components/internal/Table.tsx
EmilioMena: row.id is already used as a key above on line 84. It was causing a bunch of errors in the console. 

Also i read that keys only need to be unique with their siblings and not globably unique so this should be okay? https://reactjs.org/docs/lists-and-keys.html#keys-must-only-be-unique-among-siblings

lib/applicants/schema.graphql
EmilioMena: i tried to figure this out and it was pretty tough for an array. We could think about changing the format of a SortOptions to be an array of objects `[{field: string, order: SortOrder}]` instead of `[[string, sortOrder]]` and that would be easier to implement in gql

it looks like its a known issue https://github.com/graphql/graphql-spec/issues/423

tools/types.ts
EmilioMena: afaik gql doesn't have readonly types so when trying to assign this to the order argument in the filter whose type is
`Maybe<Array<Array<Scalars['String']>>>`
 typescript would complain

tools/components/internal/table.ts
EmilioMena: this was a side effect of removing the readonly from the other type

lib/applicants/field-resolvers.ts
OustanDing: Oh that's a great catch! I'll note this down for our backlog when we do optimizations

lib/applicants/resolvers.ts
```diff
@@ -8,14 +8,142 @@ import {
...
+  const skip = filter?.offset;
+
+  const totalCount = await prisma.applicant.count({
+    where,
+  });
+
+  const applicants = await prisma.applicant.findMany({
+    where,
+    skip: skip,
+    take: take,
```
OustanDing, line 136:
```suggestion
    skip,
    take,
```

lib/applicants/schema.graphql
OustanDing: Okay good idea, will add this investigation to backlog.

components/internal/Table.tsx
OustanDing: Interesting, that's a good catch. It didn't make any sense being in there.

components/internal/Table.tsx
OustanDing: Sounds good!

tools/pages/permit-holders/filter-permit-holders.ts
```diff
@@ -0,0 +1,54 @@
+import { gql } from '@apollo/client'; // gql tag
```
OustanDing, line 1: 1. Could we rename this file to `permit-holders-table.ts`?
2. `FilterPermitHoldersRequest` and `FilterPermitHoldersResponse` should be renamed to `GetPermitHoldersRequest` and `GetPermitHoldersResponse` respectively

OustanDing (changes requested):
2 tiny comments + 1 outstanding item from last review. Then it's good to go!

pages/admin/permit-holders.tsx
EmilioMena: mb forgot to respond to this one. The reason i have it false is bc the default for the name column is asc so clicking the column won't change anything the first time. I can remove it but i think it's better than making the user click the column twice to change order the first time

EmilioMena: the sortDescFirst option is for the first click btw the name is a bit deceiving

OustanDing (approved):
LGTM! Great work :D

PR #44, opened 2021-08-04
[Feature] Build view individual permit holder page

j-bunk:
## Notion ticket link

[View individual Permit Holder page](https://www.notion.so/uwblueprintexecs/View-individual-Permit-Holder-page-396f2b4450ee4b30bab184678d5768da)

## Implementation description
Demo:
pr_demo

## Notes
* Making guardian information editable has been deprioritized for the MVP and will be addressed post-mvp so the edit guardian  modal was not built (clicking the edit button on the guardian information card does nothing)
* Alerts that pop up at the bottom after editing should be closable, I couldn't figure out how to make it closable and just removed the close button and added a ticket to the task board.
* The way data is handled in the components will likely change to match API response data

## Checklist

### DIFF
--- a/components/internal/PermitHolderInfoCard.tsx
+++ b/components/internal/PermitHolderInfoCard.tsx
@@ -20,7 +20,7 @@ export default function PermitHolderInfoCard(props: PermitHolderInfoCardProps) {
     <GridItem
       display="flex"
       flexDirection="column"
-      alignItems={alignGridItems ? alignGridItems : 'flex-start'}
+      alignItems={alignGridItems || 'flex-start'}
       padding="20px 24px 24px"
       background="white"
       border="1px solid"
--- a/components/internal/PermitHolderInfoCard.tsx
+++ b/components/internal/PermitHolderInfoCard.tsx
@@ -6,6 +6,7 @@ type PermitHolderInfoCardProps = GridItemProps & {
   header: ReactNode;
   updated?: boolean;
   editModal?: ReactNode;
+  alignGridItems?: string;
 };
 
 /**
@@ -14,12 +15,12 @@ type PermitHolderInfoCardProps = GridItemProps & {
  * @returns custom Card.
  */
 export default function PermitHolderInfoCard(props: PermitHolderInfoCardProps) {
-  const { children, header, updated, editModal } = props;
+  const { children, header, updated, editModal, alignGridItems } = props;
   return (
     <GridItem
       display="flex"
       flexDirection="column"
-      alignItems="flex-start"
+      alignItems={alignGridItems ? alignGridItems : 'flex-start'}
       padding="20px 24px 24px"
       background="white"
       border="1px solid"
--- a/components/internal/RequestStatusBadge.tsx
+++ b/components/internal/RequestStatusBadge.tsx
@@ -15,110 +15,118 @@ export default function RequestStatusBadge({ variant }: Props) {
     switch (variant) {
       case 'COMPLETED':
         return (
-          <>
-            <Image
-              src="/assets/completed-icon.svg"
-              alt="Completed Application Icon"
-              height={12}
-              width={12}
-            />
-            <Box as="span" ml="2">
-              Completed
+          <Badge variant={variant} width="127px" height="32px" paddingX="12px" paddingY="4px">
+            <Box mx="0px" px="0px">
+              <Image
+                src="/assets/completed-icon.svg"
+                alt="Completed Application Icon"
+                height={12}
+                width={12}
+              />
+              <Box as="span" pl="8px">
+                Completed
+              </Box>
             </Box>
-          </>
+          </Badge>
         );
       case 'EXPIRING':
         return (
-          <>
-            <Image
-              src="/assets/expiring-icon.svg"
-              alt="Expiring Application Icon"
-              height={14}
-              width={14}
-            />
-            <Box as="span" ml="2">
-              Expiring &lt; 30 days
+          <Badge variant={variant} width="180px" height="32px" paddingX="12px" paddingY="4px">
+            <Box mx="0px" px="0px">
+              <Image
+                src="/assets/expiring-icon.svg"
+                alt="Expiring Application Icon"
+                height={14}
+                width={14}
+              />
+              <Box as="span" pl="8px">
+                Expiring &lt; 30 days
+              </Box>
             </Box>
-          </>
+          </Badge>
         );
       case 'EXPIRED':
         return (
-          <>
-            <Image
-              src="/assets/expired-icon.svg"
-              alt="Expired Application Icon"
-              height={15}
-              width={16}
-            />
-            <Box as="span" ml="2">
-              Expired
+          <Badge variant={variant} width="105px" height="32px" paddingX="12px" paddingY="4px">
+            <Box mx="0px" px="0px">
+              <Image
+                src="/assets/expired-icon.svg"
+                alt="Expired Application Icon"
+                height={15}
+                width={16}
+              />
+              <Box as="span" pl="8px">
+                Expired
+              </Box>
             </Box>
-          </>
+          </Badge>
         );
       case 'PENDING':
         return (
-          <>
-            <Image
-              src="/assets/pending-icon.svg"
-              alt="Pending Application Icon"
-              height={12}
-              width={12}
-            />
-            <Box as="span" ml="2">
-              Pending
+          <Badge variant={variant} width="107px" height="32px" paddingX="12px" paddingY="4px">
+            <Box mx="0px" px="0px">
+              <Image
+                src="/assets/pending-icon.svg"
+                alt="Pending Application Icon"
+                height={12}
+                width={12}
+              />
+              <Box as="span" pl="8px">
+                Pending
+              </Box>
             </Box>
-          </>
+          </Badge>
         );
       case 'REJECTED':
         return (
-          <>
-            <Image
-              src="/assets/rejected-icon.svg"
-              alt="Rejected Application Icon"
-              height={12}
-              width={12}
-            />
-            <Box as="span" ml="2">
-              Rejected
+          <Badge variant={variant} width="109px" height="32px" paddingX="12px" paddingY="4px">
+            <Box mx="0px" px="0px">
+              <Image
+                src="/assets/rejected-icon.svg"
+                alt="Rejected Application Icon"
+                height={12}
+                width={12}
+              />
+              <Box as="span" pl="8px">
+                Rejected
+              </Box>
             </Box>
-          </>
+          </Badge>
         );
       case 'INPROGRESS':
         return (
-          <>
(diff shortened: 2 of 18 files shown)


pages/admin/permit-holder/[permitHolderId].tsx
```diff
@@ -0,0 +1,122 @@
...
+
+// Individual permit holder page
+export default function PermitHolder() {
+  const { applicant, physician, guardian, applicationStatus } = mockApplication;
+
+  return (
+    <Layout>
+      <GridItem rowSpan={1} colSpan={12} marginTop={3}>
+        <PermitHolderHeader
+          applicant={applicant as unknown as Applicant}
```
Comment (j-bunk) [inline pages/admin/permit-holder/[permitHolderId].tsx:82]: Temporary fix so that the CI build can pass. Will be fixed during API hookup since the way data is handled will change based on what data is received from the APIs

components/permit-holders/GuardianInformationCard.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const { guardian } = props;
+  return (
+    <PermitHolderInfoCard
+      colSpan={7}
+      header={`Guardian's Information`}
+      editModal={
+        //   TODO: Create edit guardian modal post-mvp
+        <Button color="primary" variant="ghost" textDecoration="underline">
+          <Text textStyle="body-bold">Edit</Text>
+        </Button>
```
angeladietz, line 19: I think we should just remove this button temporarily since it doesn't do anything

components/permit-holders/AppHistoryCard.tsx
```diff
@@ -0,0 +1,111 @@
...
+    </Link>
+  );
+}
+
+export default function AppHistoryCard() {
+  return (
+    <PermitHolderInfoCard header={`APP History`} alignGridItems="normal">
+      <Divider pt="24px" />
+      <Box padding="20px 24px">
+        <Table columns={COLUMNS} data={DATA} />
```
angeladietz, line 107: For this table as well as the ones in Attached Files and Medical History, there's no divider between the header and data. I personally kinda like the divider line and it would also be more simple to leave it since they use the reusable table component, but we should get a designer's input to see if this is okay

components/permit-holders/MedicalHistoryCard.tsx
```diff
@@ -0,0 +1,69 @@
...
+  disablingCondition: 'Condition 1',
+  associatedApp: { associatedApp: 12345 },
+  dateUploaded: '2021/01/01',
+  fileUrl: { fileUrl: '/' },
+  application: { application: mockMedicalHistory },
+});
+
+const COLUMNS = [
+  {
+    Header: 'Disabling Condition(s)',
```
angeladietz, line 28: (more of a design critique but) I think we should change "Disabling Condition(s)" to "Disabling Condition". Unless multiple conditions will go in the same row, we don't need it, and even in that case it still seems unnecessary to have that "(s)" there.

components/permit-holders/MedicalHistoryCard.tsx
```diff
@@ -0,0 +1,69 @@
...
+    accessor: 'disablingCondition',
+    disableSortBy: true,
+    minWidth: 250,
+  },
+  {
+    Header: 'Certification Date',
+    accessor: 'dateUploaded',
+    disableSortBy: true,
+    // NOTE: Check value
```
angeladietz, line 37: If this is a TODO for future dev work then I think you should add more context. Otherwise if you could resolve it that would be great

components/permit-holders/AppHistoryCard.tsx
```diff
@@ -0,0 +1,111 @@
...
+function _renderRequestType({ value }: RequestTypeProps) {
+  return (
+    <>
+      <Text as="p">{value.isRenewal ? `Renewal` : `Replacement`}</Text>
+    </>
+  );
+}
+
+type StatusProps = {
+  value: 'COMPLETED' | 'INPROGRESS' | 'PENDING' | 'REJECTED' | 'EXPIRING' | 'EXPIRED' | 'ACTIVE';
```
angeladietz, line 77: There's now an `ApplicationStatus` enum which was added by Oustan recently so it would be great if we could use that here instead. I know it needs to still be fixed in other places too so if its not trivial to only use it here then maybe don't worry about it and it can be fixed separately.

components/permit-holders/AppHistoryCard.tsx
```diff
@@ -0,0 +1,111 @@
...
+    accessor: 'isRenewal',
+    disableSortBy: true,
+    maxWidth: 140,
+    Cell: _renderRequestType,
+  },
+  {
+    Header: 'Status',
+    accessor: 'requestStatus',
+    disableSortBy: true,
+    maxWidth: 190,
```
angeladietz, line 35: I think the status column seems too narrow on my screen, and I'm using a MacBook Air which has a fairly standard screen size. Is there any way the column widths could be adjusted to give more space between the status column and expiry date?
Screen Shot 2021-08-07 at 11 12 40 PM

components/permit-holders/PermitHolderHeader.tsx
```diff
@@ -0,0 +1,42 @@
+import { Box, Flex, HStack, Text } from '@chakra-ui/react'; // Chakra UI
+import { ChevronLeftIcon } from '@chakra-ui/icons'; // Chakra UI icon
+import Link from 'next/link'; // Link
+import RequestStatusBadge from '@components/internal/RequestStatusBadge'; // Request status badge
+import { Applicant } from '@lib/graphql/types'; // Applicant type
+
+type PermitHolderHeaderProps = {
+  readonly applicant: Applicant;
+  readonly applicationStatus: // TODO: Change this to the enum that we add
```
angeladietz, line 9: This enum exists now (although I don't think it has all of these values... )

components/internal/PermitHolderInfoCard.tsx
```diff
@@ -14,12 +15,12 @@ type PermitHolderInfoCardProps = GridItemProps & {
...
  */
 export default function PermitHolderInfoCard(props: PermitHolderInfoCardProps) {
-  const { children, header, updated, editModal } = props;
+  const { children, header, updated, editModal, alignGridItems } = props;
   return (
     <GridItem
       display="flex"
       flexDirection="column"
-      alignItems="flex-start"
+      alignItems={alignGridItems ? alignGridItems : 'flex-start'}
```
angeladietz, line 23: I think we can simplify this to: ` alignItems={alignGridItems || 'flex-start'}`

angeladietz (changes requested):
Looks awesome! Just a couple small suggestions.

components/permit-holders/MedicalHistoryCard.tsx
j-bunk: Just a note to myself that I forgot to remove, nice catch!

components/permit-holders/AppHistoryCard.tsx
j-bunk: Agreed looks a bit too narrow on laptop screens, added 10px of padding to the right to address this
image

components/permit-holders/AppHistoryCard.tsx
j-bunk: The enum Oustan added has different values than the values here (enum values: PENDING, APPROVED, REJECTED, COMPLETED). I think APPROVED == INPROGRESS and the rest are calculated based on the expiry date. 🤔 I'm not entirely sure how to handle that atm, i think i might just address it separately later

j-bunk: cc @OustanDing

components/permit-holders/PermitHolderHeader.tsx
j-bunk: Related comment: https://github.com/uwblueprint/richmond-centre-for-disability/pull/44#discussion_r684750450

components/internal/RequestStatusBadge.tsx
```diff
@@ -15,110 +15,118 @@ export default function RequestStatusBadge({ variant }: Props) {
...
-          <>
-            <Image
-              src="/assets/completed-icon.svg"
-              alt="Completed Application Icon"
-              height={12}
-              width={12}
-            />
-            <Box as="span" ml="2">
-              Completed
+          <Badge variant={variant} width="127px" height="32px" paddingX="12px" paddingY="4px">
```
j-bunk, line 18: The `RequestStatusBadge` seem to sometimes resize based on the container they're in so I refactored it and specified the height and width explicitly

components/permit-holders/AppHistoryCard.tsx
j-bunk: Yep good idea! I'll message the designers

components/permit-holders/MedicalHistoryCard.tsx
j-bunk: So medical history shows the permit holder's medical conditions at the time of that application. Meaning the first row is all their medical conditions at the time of the most recent application, and the second row is all their medical conditions at the time of the application prior. So multiple conditions can go into one row, but I do agree though that the "(s)" isn't the most aesthetically pleasing. I'll bring it up with the designers and get their input

j-bunk:
@angeladietz Thanks for all the great feedback! Addressed most of it and left comments for the ones I haven't fully addressed. I'll also message the designers to get their inputs on the points you brought up

components/internal/RequestStatusBadge.tsx
angeladietz: Good catch! I think a simpler fix for this might have been to leave this file unchanged, except change the last return statement to this:

```
  return (
    <Wrap>
      <Badge variant={variant}>
        <Box ml="-4px" mr="8px">
          {_renderBadgeContent(variant)}
        </Box>
      </Badge>
    </Wrap>
  );
```

The only difference between this and your original code is the `<Wrap></Wrap>`, and this seemed to fix the issue as well. Feel free to stick to either solution, although this one would eliminate some of those hardcoded widths.

components/permit-holders/AppHistoryCard.tsx
j-bunk: Andy said it's we're all good to just keep the divider for now

components/permit-holders/MedicalHistoryCard.tsx
j-bunk: Dropped the "(s)" after getting Andy's input + approval

components/internal/RequestStatusBadge.tsx
j-bunk: That looks cleaner so I implemented your suggestion!

I noticed that some positioning/spacing was slightly off so I tweaked the padding a bit to match the designs more closely
Before:
image

After minor padding tweaks (also modified one line from `badge.ts`):
Expiring  30 days

angeladietz (approved):
LGTM 🚀

PR #45, opened 2021-08-05
[Feature] Request Table API Hookup

angeladietz:
## Notion ticket link

[Requests API hookup](https://www.notion.so/uwblueprintexecs/Requests-API-hookup-ec81141ad07c48dbaf8bf1287f687420)

## Implementation description
* Hooks up API for request table to display appropriate applications
* supports filtering, pagination, sorting, searching
* added `permitType` field to `Applications` DB table
* created `PermitType` enum
* Added more values to `ApplicationStatus` enum

## Notes
* We might need to fix a small UX bug with the sorting -- the arrow disappears once the data refreshes and will only reappear if we click somewhere on the header
* This PR will also probably conflict with #42 

## Checklist

### DIFF
--- a/lib/applications/resolvers.ts
+++ b/lib/applications/resolvers.ts
@@ -11,10 +11,10 @@ import { DBErrorCode } from '@lib/db/errors'; // Database errors
  * Query all the RCD applications in the internal-facing app
  * @returns All RCD applications
  */
-export const applications: Resolver = async (_parent, args, { prisma }) => {
+export const applications: Resolver = async (_parent, { filter }, { prisma }) => {
   const {
     input: { order, status, search, limit = 20, offset = 0 }, // TODO: permitType, requestType,
-  } = args;
+  } = filter;
 
   let userIDSearch, nameSearch;
 
--- a/lib/applications/resolvers.ts
+++ b/lib/applications/resolvers.ts
@@ -11,17 +11,22 @@ import { DBErrorCode } from '@lib/db/errors'; // Database errors
  * Query all the RCD applications in the internal-facing app
  * @returns All RCD applications
  */
-// export const applications: Resolver = async (_parent, _args, { prisma }) => {
-//   const {
-//     input: { order, status, permitType, requestType, search, limit=20, offset=0 },
-//   } = args;
-//   const applications = await prisma.application.findMany({
-//     where: {
+export const applications: Resolver = async (_parent, _args, { prisma }) => {
+  const {
+    input: { order, limit = 20, offset = 0 },
+  } = _args;
 
-//     }
-//   });
-//   return applications;
-// };
+  const sortingOrder: Record<string, string> = order
+    ? order.foreach((col: string, order: string) => (sortingOrder[col] = order))
+    : { dateReceived: 'desc' };
+
+  const applications = await prisma.application.findMany({
+    skip: offset,
+    take: limit,
+    orderBy: [sortingOrder],
+  });
+  return applications;
+};
 
 /**
  * Create an RCD application
--- a/lib/applications/resolvers.ts
+++ b/lib/applications/resolvers.ts
@@ -11,10 +11,17 @@ import { DBErrorCode } from '@lib/db/errors'; // Database errors
  * Query all the RCD applications in the internal-facing app
  * @returns All RCD applications
  */
-export const applications: Resolver = async (_parent, _args, { prisma }) => {
-  const applications = await prisma.application.findMany();
-  return applications;
-};
+// export const applications: Resolver = async (_parent, _args, { prisma }) => {
+//   const {
+//     input: { order, status, permitType, requestType, search, limit=20, offset=0 },
+//   } = args;
+//   const applications = await prisma.application.findMany({
+//     where: {
+
+//     }
+//   });
+//   return applications;
+// };
 
 /**
  * Create an RCD application
--- a/lib/applications/resolvers.ts
+++ b/lib/applications/resolvers.ts
@@ -9,99 +9,97 @@ import { DBErrorCode } from '@lib/db/errors'; // Database errors
 import { SortOrder } from '@tools/types'; // Sorting type
 
 /**
- * Query all the RCD applications in the internal-facing app
- * @returns All RCD applications
+ * Query and filter RCD applications from the internal facing app.
+ * All fields are optional.
+ *
+ * Sorting:
+ * - order: array of tuples of the field being sorted and the order. Default [['dateReceived', 'desc']]
+ *
+ * Filters:
+ * - permitType: PermitType (PERMANENT, TEMPORARY)
+ * - requestType: Replacement, Renewal
+ * - status: ApplicationStatus(PENDING, INPROGRESS, APPROVED, REJECTED, COMPLETED, EXPIRING, EXPIRED, ACTIVE)
+ * - search: Search by first, middle, last name or by RCD user ID
+ *
+ * Pagination:
+ * - limit: Number of result to return
+ * - offset: Number of results to skip
+ *
+ * @returns All RCD applications that match the filter(s).
  */
 export const applications: Resolver = async (_parent, { filter }, { prisma }) => {
-  const { order, permitType, requestType, status, search, limit = 20, offset = 0 } = filter;
-
-  // Parse search string
-  let userIDSearch, firstSearch, middleSearch, lastSearch;
-
-  if (parseInt(search)) {
-    userIDSearch = parseInt(search);
-  } else if (search) {
-    [firstSearch, middleSearch, lastSearch] = search.split(' ');
-    middleSearch = middleSearch || firstSearch;
-    lastSearch = lastSearch || middleSearch;
-  }
-
-  // Parse sorting order
+  let where = {};
   let orderBy = undefined;
 
-  if (order && order.length > 0) {
-    const sortingOrder: Array<Record<string, SortOrder>> = [];
-    order.forEach(([field, order]: [string, SortOrder]) => {
-      if (field === 'name') {
-        // Primary sort is by first name and secondary sort is by last name
-        sortingOrder.push({ firstName: order });
-        sortingOrder.push({ lastName: order });
-      } else if (field === 'dateReceived') {
-        sortingOrder.push({ createdAt: order });
-      }
-    });
-    orderBy = sortingOrder;
+  if (filter) {
+    const {
+      order = undefined,
+      permitType = undefined,
+      requestType = undefined,
+      status = undefined,
+      search = undefined,
+    } = filter;
+
+    // Parse search string
+    let userIDSearch, firstSearch, middleSearch, lastSearch;
+
+    if (parseInt(search)) {
+      userIDSearch = parseInt(search);
+    } else if (search) {
+      [firstSearch, middleSearch, lastSearch] = search.split(' ');
+      middleSearch = middleSearch || firstSearch;
+      lastSearch = lastSearch || middleSearch;
+    }
+
+    // Parse sorting order
+    if (order && order.length > 0) {
+      const sortingOrder: Array<Record<string, SortOrder>> = [];
+      order.forEach(([field, order]: [string, SortOrder]) => {
+        if (field === 'name') {
+          // Primary sort is by first name and secondary sort is by last name
+          sortingOrder.push({ firstName: order });
+          sortingOrder.push({ lastName: order });
+        } else if (field === 'dateReceived') {
+          sortingOrder.push({ createdAt: order });
+        }
+      });
+      orderBy = sortingOrder;
+    }
+
+    where = {
+      applicant: {
+        id: userIDSearch,
+      },
+      applicationProcessing: {
+        status: status || undefined,
+      },
+      isRenewal: requestType ? requestType === 'Renewal' : undefined,
+      permitType: permitType || undefined,
+      AND: [
+        {
+          OR: [
+            { firstName: { contains: firstSearch, mode: 'insensitive' } },
+            { middleName: { contains: middleSearch, mode: 'insensitive' } },
+            { lastName: { contains: lastSearch, mode: 'insensitive' } },
+          ],
+        },
+      ],
+    };
   }
 
   // Get number of applications with desired filters
   const applicationsCount = await prisma.application.count({
-    where: {
-      applicant: {
-        id: userIDSearch,
-      },
-      applicationProcessing: {
-        status: status || undefined,
-      },
-      isRenewal: requestType ? requestType === 'Renewal' : undefined,
-      permitType: permitType || undefined,
-      AND: [
-        {
-          OR: [
-            { firstName: { contains: firstSearch, mode: 'insensitive' } },
-            { middleName: { contains: middleSearch, mode: 'insensitive' } },
-            { lastName: { contains: lastSearch, mode: 'insensitive' } },
-          ],
-        },
-      ],
-    },
+    where: where,
   });
 
   // Get applications with filter, sorting, pagination
   const applications = await prisma.application.findMany({
-    skip: offset,
-    take: limit,
+    skip: filter?.offset || 0,
+    take: filter?.limit || 20,
     orderBy: orderBy,
-    where: {
-      applicant: {
-        id: userIDSearch,
-      },
-      applicationProcessing: {
-        status: status || undefined,
-      },
-      isRenewal: requestType ? requestType === 'Renewal' : undefined,
-      permitType: permitType || undefined,
-      AND: [
-        {
-          OR: [
-            { firstName: { contains: firstSearch, mode: 'insensitive' } },
-            { middleName: { contains: middleSearch, mode: 'insensitive' } },
-            { lastName: { contains: lastSearch, mode: 'insensitive' } },
-          ],
-        },
-      ],
-    },
-    select: {
-      firstName: true,
-      lastName: true,
-      id: true,
-      createdAt: true,
-      permitType: true,
-      isRenewal: true,
-      applicationProcessing: {
-        select: {
-          status: true,
-        },
-      },
+    where: where,
+    include: {
+      applicationProcessing: true,
     },
   });
 
--- a/lib/applications/resolvers.ts
+++ b/lib/applications/resolvers.ts
@@ -71,10 +71,10 @@ export const applications: Resolver = async (_parent, { filter }, { prisma }) =>
         id: userIDSearch,
       },
       applicationProcessing: {
-        status: status || undefined,
+        status: status,
       },
       isRenewal: requestType ? requestType === 'Renewal' : undefined,
-      permitType: permitType || undefined,
+      permitType: permitType,
       AND: [
         {
           OR: [
--- a/lib/applications/resolvers.ts
+++ b/lib/applications/resolvers.ts
@@ -89,7 +89,7 @@ export const applications: Resolver = async (_parent, { filter }, { prisma }) =>
 
   // Get number of applications with desired filters
   const applicationsCount = await prisma.application.count({
-    where: where,
+    where,
   });
 
   // Get applications with filter, sorting, pagination
@@ -97,7 +97,7 @@ export const applications: Resolver = async (_parent, { filter }, { prisma }) =>
     skip: filter?.offset || 0,
     take: filter?.limit || 20,
     orderBy: orderBy,
-    where: where,
+    where,
     include: {
       applicationProcessing: true,
     },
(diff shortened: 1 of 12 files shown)


lib/applications/resolvers.ts
```diff
@@ -6,14 +6,109 @@ import {
...
+      AND: [
+        {
+          OR: [
+            { firstName: { contains: firstSearch, mode: 'insensitive' } },
+            { middleName: { contains: middleSearch, mode: 'insensitive' } },
+            { lastName: { contains: lastSearch, mode: 'insensitive' } },
+          ],
+        },
+      ],
+    },
```
EmilioMena, line 66: So you don't have to repeat this in both the count query and actual query you can take out the `where` in a variable and pass it to both queries. it will also be easier to make a change in one place in the future if needed

lib/applications/resolvers.ts
```diff
@@ -6,14 +6,109 @@ import {
...
+    select: {
+      firstName: true,
+      lastName: true,
+      id: true,
+      createdAt: true,
+      permitType: true,
+      isRenewal: true,
+      applicationProcessing: {
+        select: {
+          status: true,
```
EmilioMena, line 102: Is there a reason we're explicitly selecting only these columns? I don't think we should constrain which fields are being returned. Graphql already handles selecting only the data that we need in the query.

lib/applications/schema.graphql
```diff
@@ -124,3 +126,43 @@ input CreateApplicationInput {
...
+
+type ApplicationsFilterResult{
+  firstName: String!
+  lastName: String!
+  id: ID!
+  createdAt: Date!
+  applicationProcessing: ApplicationProcessing
+  isRenewal: Boolean!
+  permitType: PermitType!
+}
```
EmilioMena, line 163: I don't think we need a custom return type when filtering applications, `Application` should work just fine, no?
applicationProcessing should also probably be added to the Application type as well

pages/admin/index.tsx
```diff
@@ -57,15 +60,16 @@ function renderName({ value }: NameProps) {
   return (
     <div>
       <Text>{value.name}</Text>
-      <Text textStyle="caption" textColor="secondary">
-        ID: {value.id}
-      </Text>
+      {value.id && (
```
EmilioMena, line 63: hm shouldn't there always be an ID for an application?

pages/admin/index.tsx
```diff
@@ -104,44 +112,73 @@ const COLUMNS = [
...
+          offset: pageNumber * PAGE_SIZE,
+          limit: PAGE_SIZE,
+        },
+      },
+      notifyOnNetworkStatusChange: true,
+      onCompleted: data => {
+        setRequestsData(
+          data.applications.result.map(record => ({
+            name: {
+              name: record.firstName + ' ' + record.lastName,
```
EmilioMena, line 163: is middle name also needed here?

pages/admin/index.tsx
```diff
@@ -104,44 +112,73 @@ const COLUMNS = [
...
+              rcdUserId: record.id,
```
EmilioMena, line 164: ah it looks like we're using the database id and not the rcdUserId here, probably need to change that

pages/admin/index.tsx
```diff
@@ -104,44 +112,73 @@ const COLUMNS = [
...
+    rcdUserId: number;
+  };
+  dateReceived: Date;
+  permitType: string;
+  requestType: string;
+  status: ApplicationStatus | undefined;
+};
+
+// Max number of entries in a page
+const PAGE_SIZE = 1;
```
EmilioMena, line 128: I had this as 1 for testing too lol just change it back to 20 before merging

EmilioMena (commented):
Awesome work so far y'all! Left down some comments

lib/applications/resolvers.ts
angeladietz: Good point. we had to specifically select the applicationProcessing field but I realized we can use `include` instead so I switched to that

lib/applications/schema.graphql
angeladietz: Yeah you're right, good catch

pages/admin/index.tsx
angeladietz: The ID is the applicant ID, not the application ID, so if someone is applying for the first time, they won't have an applicant ID yet.

pages/admin/index.tsx
angeladietz: We're not showing the middle name on the table (which follows the design) so it's left out.

pages/admin/index.tsx
```diff
@@ -266,13 +349,32 @@ export default function Requests() {
...
-                  <Input placeholder="Search by first name, last name or user ID" />
+                  <Input
+                    placeholder="Search by name or user ID"
+                    onKeyDown={event => {
+                      if (event.key === 'Enter') {
+                        setSearchFilter(searchInput);
+                        setPageNumber(0);
+                      }
+                    }}
+                    onChange={event => setSearchInput(event.target.value)}
```
EmilioMena, line 360: Oustan and I added a [useDebounce](https://usehooks.com/useDebounce/) hook in my [latest commit](https://github.com/uwblueprint/richmond-centre-for-disability/pull/42/commits/94abb4044c7b17244d7adf3e8532ce6983f8a9af) so that we only need to use onChange here. The debouncer will only set the searchFilter state if it's value hasn't been changed in the last 500ms. That way we can avoid making users press enter to search without firing a db query every time a key is pressed. It's better ux and looks cooler imo

Can you copy those changes in here?

pages/admin/index.tsx
```diff
@@ -266,13 +349,32 @@ export default function Requests() {
...
+                onChangeSortOrder={sortOrder => setSortOrder(sortOrder)}
+              />
+            )}
             <Flex justifyContent="flex-end">
-              <Pagination pageSize={20} totalCount={150} />
+              <Pagination
+                pageSize={PAGE_SIZE}
+                totalCount={recordsCount}
+                onPageChange={n => setPageNumber(n)}
+              />
```
EmilioMena, line 377: also fixed a bug in the pagination component in my latest commit. It now sets the page number in the props instead of keeping the logic inside the component which allows the pageNumber to be set to 0 after a filter change

pages/admin/index.tsx
```diff
@@ -234,27 +314,30 @@ export default function Requests() {
...
+                      setRequestTypeFilter(undefined);
+                      setPageNumber(0);
                     }}
                   >
                     All
                   </MenuItem>
                   <MenuItem
                     onClick={() => {
                       setRequestTypeFilter('Replacement');
+                      setPageNumber(0);
```
EmilioMena, line 332: can you group all the setPageNumber(0) in a useEffect hook? That it doesn't have to be repeated for every filter

pages/admin/index.tsx
```diff
@@ -104,44 +117,82 @@ const COLUMNS = [
...
+  // Map uppercase enum strings to lowercase
+  const permitTypeString: Record<PermitType, string> = {
+    [PermitType.Permanent]: 'Permanent',
+    [PermitType.Temporary]: 'Temporary',
+  };
+
+  // Set page number to 0 after every filter or sort change
+  useEffect(() => {
+    setPageNumber(0);
+  }, [statusFilter, permitTypeFilter, requestTypeFilter, debouncedSearchFilter, sortOrder]);
```
EmilioMena, line 195: Do we need sortOrder here?

tools/pages/admin/index.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+`;
+
+export type GetApplicationsRequest = {
+  filter: ApplicationsFilter;
+};
+
+export type GetApplicationsResponse = {
+  applications: {
+    // result: [Application];
```
EmilioMena, line 30: remove comment

EmilioMena (approved):
Awesome job! Left a couple small questions/comments but other than that good to go 🚢

pages/admin/index.tsx
angeladietz: Yeah we do assuming that when the sorting order changes we want to go back to the first page (which I assume is the case because the data on the current page will most likely be different).

lib/applications/resolvers.ts
```diff
@@ -6,14 +6,107 @@ import {
...
+ * - permitType: PermitType (PERMANENT, TEMPORARY)
+ * - requestType: Replacement, Renewal
+ * - status: ApplicationStatus(PENDING, INPROGRESS, APPROVED, REJECTED, COMPLETED, EXPIRING, EXPIRED, ACTIVE)
+ * - search: Search by first, middle, last name or by RCD user ID
+ *
+ * Pagination:
+ * - limit: Number of result to return
+ * - offset: Number of results to skip
+ *
+ * @returns All RCD applications that match the filter(s).
```
OustanDing, line 28: Nice documentation 🤩

lib/applications/resolvers.ts
```diff
@@ -6,14 +6,107 @@ import {
...
+    } else if (search) {
+      [firstSearch, middleSearch, lastSearch] = search.split(' ');
+      middleSearch = middleSearch || firstSearch;
+      lastSearch = lastSearch || middleSearch;
+    }
+
+    // Parse sorting order
+    if (order && order.length > 0) {
+      const sortingOrder: Array<Record<string, SortOrder>> = [];
+      order.forEach(([field, order]: [string, SortOrder]) => {
```
OustanDing, line 57: Can use `SortOptions` from `tools/types.ts`

lib/applications/schema.graphql
```diff
@@ -60,6 +62,8 @@ type Application {
 
   # Permit
   permit: Permit
+
+  createdAt: Date
```
OustanDing, line 66:
```suggestion
  createdAt: Date!
```

lib/applications/schema.graphql
```diff
@@ -124,3 +129,33 @@ input CreateApplicationInput {
...
+  appNumber: Int
+  appHolepunched: Boolean
+  walletCardCreated: Boolean
+  invoiceNumber: Int
+  documentUrls: [String]
+  appMailed: Boolean
+  applicationId: Int
+  createdAt: Date
+  updatedAt: Date
+  application: Application
```
OustanDing, line 155:
```suggestion
  id: Int!
  status: ApplicationStatus!
  appNumber: Int
  appHolepunched: Boolean!
  walletCardCreated: Boolean!
  invoiceNumber: Int
  documentUrls: [String!]
  appMailed: Boolean!
  applicationId: Int!
  createdAt: Date!
  updatedAt: Date!
  application: Application!
```

lib/graphql/schema.graphql
```diff
@@ -100,4 +104,9 @@ enum PermitStatus {
 enum UserStatus {
   ACTIVE
   INACTIVE
-}
\ No newline at end of file
+}
+
+enum  PermitType {
```
OustanDing, line 109:
```suggestion
enum PermitType {
```

lib/graphql/schema.graphql
```diff
@@ -80,9 +80,13 @@ enum Gender {
 
 enum ApplicationStatus {
   PENDING
+  INPROGRESS
   APPROVED
   REJECTED
   COMPLETED
+  EXPIRING
+  EXPIRED
+  ACTIVE
```
OustanDing, line 89: This enum is used for the application processing steps, so it should only contain `PENDING`, `APPROVED`, `REJECTED` and `COMPLETED`.

`EXPIRING`, `EXPIRED` and `ACTIVE` are computed dynamically. `INPROGRESS` is equal to `APPROVED` (if an application is approved but not yet completed, it is currently being processed).

pages/admin/index.tsx
```diff
@@ -49,23 +57,24 @@ function renderStatusBadge({ value }: StatusProps) {
 type NameProps = {
   readonly value: {
     name: string;
-    id: string;
+    rcdUserId: string;
   };
 };
 
 function renderName({ value }: NameProps) {
```
OustanDing, line 64: Similar to in @EmilioMena 's PR, I think we can just put all of these render functions in the columns configuration to keep it all in one place. Also, I believe it can infer the types if you do it like that.

tools/pages/admin/index.ts
```diff
@@ -0,0 +1,45 @@
...
+        | 'firstName'
+        | 'lastName'
+        | 'id'
+        | 'createdAt'
+        | 'permitType'
+        | 'isRenewal'
+        | 'applicantId'
+        | 'applicationProcessing'
+      >
+    >;
```
OustanDing, line 42:
```suggestion
    result: ReadonlyArray<
      Pick<
        Application,
        | 'firstName'
        | 'lastName'
        | 'id'
        | 'createdAt'
        | 'permitType'
        | 'isRenewal'
        | 'applicantId'
      > & {
        applicationProcessing: Pick<ApplicationProcessing, 'status'>
      }
    >;
```
We also want to pick from the `applicationProcessing` object

pages/admin/index.tsx
OustanDing: Agreed

pages/admin/index.tsx
```diff
@@ -104,44 +117,82 @@ const COLUMNS = [
...
+
+  // Sorting
+  const [sortOrder, setSortOrder] = useState<SortOptions>([['dateReceived', SortOrder.DESC]]);
+
+  // Debounce search filter so that it only gives us latest value if searchFilter has not been updated within last 500ms.
+  // This will avoid firing a query for each key the user presses
+  const debouncedSearchFilter = useDebounce<string>(searchFilter, 500);
+
+  // Data & pagination
+  const [requestsData, setRequestsData] = useState<ApplicationData[]>();
```
OustanDing, line 151: We don't need a state variable for this -  the page and table will rerender when any of the filters change, so the table data will be updated automatically as well.

pages/admin/index.tsx
```diff
@@ -104,44 +117,82 @@ const COLUMNS = [
...
+  const [pageNumber, setPageNumber] = useState(0);
+  const [recordsCount, setRecordsCount] = useState(0);
```
OustanDing, line 153: Also don't need a state variable for `recordsCount`

pages/admin/index.tsx
```diff
@@ -104,44 +117,82 @@ const COLUMNS = [
...
+
+  // Make query to applications resolver
+  useQuery<GetApplicationsResponse, GetApplicationsRequest>(GET_APPLICATIONS_QUERY, {
```
OustanDing, line 156: Following the two comments above, we can grab the query result like this:
```suggestion
  const { data } = useQuery<GetApplicationsResponse, GetApplicationsRequest>(GET_APPLICATIONS_QUERY, {
```
and we can destructure the `results` and `totalCount` from `data`

pages/admin/index.tsx
```diff
@@ -104,44 +117,82 @@ const COLUMNS = [
...
+        data.applications.result.map(record => ({
+          name: {
+            name: record.firstName + ' ' + record.lastName,
+            rcdUserId: record.applicantId || undefined,
+          },
+          dateReceived: record.createdAt,
+          permitType: permitTypeString[record.permitType],
+          requestType: record.isRenewal ? 'Renewal' : 'Replacement',
+          status: record.applicationProcessing?.status || undefined,
+        }))
```
OustanDing, line 180: We want to move this table data formatting logic to the `Cell` parameter in the column configuration just to keep everything in one place.

OustanDing (changes requested):
Great work! Left some comments, mostly regarding consistency. Functionality and logic look sound. After that, we should be good to merge :)

lib/applications/resolvers.ts
jennifertsai: If we use the `SortOptions` instead, we can't pass it into Prisma directly for the `orderBy` field

angeladietz: ^ yeah, `order` is already a `SortOptions` so here we're parsing it into a format that we can pass into the prisma query.

pages/admin/index.tsx
angeladietz: We actually can't do this because the data value is never used. The `data` that's used inside onCompleted doesn't correspond to the same data variable so it produces errors

lib/applications/resolvers.ts
OustanDing: Gotcha, thanks!

pages/admin/index.tsx
```diff
@@ -16,61 +16,53 @@ import {
...
-    </Wrap>
-  );
-}
-
-type NameProps = {
-  readonly value: {
-    name: string;
-    id: string;
-  };
+type PermitTypeProps = {
```
OustanDing, line 45: Tiny nit. Name is misleading because these aren't component props. Easiest fix would be to just remove this type.

pages/admin/index.tsx
```diff
@@ -16,61 +16,53 @@ import {
...
-const COLUMNS = [
+const COLUMNS: Column<any>[] = [
   {
     Header: 'Name',
     accessor: 'name',
-    Cell: renderName,
+    Cell: ({ value }) => {
+      return (
+        <div>
+          <Text>{value.firstName + ' ' + value.lastName}</Text>
```
OustanDing, line 57:
```suggestion
          <Text>{`${value.firstName} ${value.lastName}`}</Text>
```
JS string interpolation

pages/admin/index.tsx
```diff
@@ -79,69 +71,119 @@ const COLUMNS = [
...
-    requestType: 'Replacement',
-    status: 'COMPLETED',
-  },
-];
+// Application data for table
+type ApplicationData = {
+  name: {
+    firstName: string;
+    lastName: string;
+    rcdUserId: number | undefined;
```
OustanDing, line 120:
```suggestion
    rcdUserId?: number;
```

pages/admin/index.tsx
```diff
@@ -79,69 +71,119 @@ const COLUMNS = [
...
+  };
+  dateReceived: Date;
+  permitType: PermitType;
+  isRenewal: boolean;
+  status: ApplicationStatus | undefined;
```
OustanDing, line 125:
```suggestion
  status: ApplicationStatus?;
```

pages/admin/index.tsx
```diff
@@ -79,69 +71,119 @@ const COLUMNS = [
...
+        data.applications.result.map(record => ({
+          name: {
+            firstName: record.firstName,
+            lastName: record.lastName,
+            rcdUserId: record.applicantId || undefined,
+          },
+          dateReceived: record.createdAt,
+          permitType: record.permitType,
+          isRenewal: record.isRenewal,
+          status: record.applicationProcessing?.status || undefined,
```
OustanDing, line 176:
```suggestion
          status: record.applicationProcessing?.status,
```
The `applicationProcessing?` should cover the `undefined` case

pages/admin/index.tsx
OustanDing: Note for future: Discussed with @angeladietz and we will keep using the state variable as it's necessary for table rerendering. My bad :/

OustanDing (commented):
Few small nits. Sorry I missed them in the previous review

pages/admin/index.tsx
```diff
@@ -37,15 +37,11 @@ import useDebounce from '@tools/hooks/useDebounce'; // Debounce hook
 import { Column } from 'react-table';
 
 // Map uppercase enum strings to lowercase
-const permitTypeString: Record<PermitType, string> = {
+const permitTypeString: Record<string, string> = {
```
OustanDing, line 40: Just wondering why this was changed? I think it should be working fine

OustanDing (approved):
Fixes look good! I just left one question in the review. Will approve since everything else looks great. Thanks so much for your great work @angeladietz @jennifertsai !

pages/admin/index.tsx
angeladietz: Without using the `PermitTypeProps`, it interprets the result as an `any` and doesn't allow for reading from `permitTypeString` since it might not be a `PermitType`.
[Exhibit A](https://railway.app/project/afc837ea-aa06-4d86-8732-287740d12581/deployments?id=3b3e73a7-e9cc-4828-8ee5-e2d600b5f5c0)
[Exhibit B](https://railway.app/project/afc837ea-aa06-4d86-8732-287740d12581/deployments?id=946771c2-4188-4a4c-97bf-a6cd156f95e0)
[Exhibit C](https://railway.app/project/afc837ea-aa06-4d86-8732-287740d12581/deployments?id=1f4921c8-40d4-4a76-a3dd-988014a4469c)

If you have a better fix then lmk and we can address it separately

OustanDing: Got it. Will add to backlog.

PR #47, opened 2021-08-06
[Bugfix] Move foreign keys to applicant model

OustanDing:
## Notion ticket link

[Ticket](https://www.notion.so/uwblueprintexecs/Move-foreign-key-from-MedicalInformation-and-Guardian-models-to-Applicant-model-732759707ca24665b69e918c4aab5077)

## Implementation description
* Moved foreign keys in `applicant-medicalInformation` and `applicant-guardian` relations to the `applicant` model to enforce existence of medical information and guardian records for each applicant
* Modularized seeding script to reduce clutter in main seeding script file
* Fixed broken APIs resulting from foreign key change

## Notes
* Tested the broken APIs again - they work. I don't expect any other places to break, since they don't depend on the relation between `applicant` and `medicalInformation`/`guardian`

## Checklist


components/requests/PersonalInformationCard.tsx
```diff
@@ -1,10 +1,12 @@
...
 import PermitHolderInfoCard from '@components/internal/PermitHolderInfoCard'; // Custom Card component
-import { Applicant } from '@lib/graphql/types'; // Applicant type
+// TODO: Temporary fix - add back in API hookup
+// import { Applicant } from '@lib/graphql/types'; // Applicant type
 import { MouseEventHandler } from 'react'; // React
 
 type PersonalInformationProps = {
-  readonly applicant: Applicant;
+  // TODO: Temporary fix - will be addressed in API hookup
+  readonly applicant: any;
```
EmilioMena, line 9: just curious why is this broken after the foreign key change

Is it just bc of the mock data?

pages/admin/request/[requestId].tsx
```diff
@@ -150,6 +150,7 @@ export default function Request() {
       </GridItem>
       <GridItem rowSpan={12} colSpan={5} marginTop={5} textAlign="left">
         <PersonalInformationCard
+          // TODO: Temporary fix - will be addressed in API hookup
```
Comment (EmilioMena) [inline pages/admin/request/[requestId].tsx:153]: what's the fix here, don't see any changes in this file?

EmilioMena (approved):
great job catching this early on, could've caused some major headaches down the line 🚀 
also nice work with cleaning up the seed file

this looks good as is, just left a couple of questions

components/requests/PersonalInformationCard.tsx
OustanDing: Good question! Yeah, it's because the mock data no longer fits the `Applicant` type. Adding the foreign keys to the `applicant` model makes the `Applicant` type also contain the nested `guardian` and `medicalInformation` objects - the workaround is to use TypeScript's `Pick` type to just include the fields that are applicable to the View Request page. cc @jeffreyzhang2001 who will address this in Request page API hookup.

pages/admin/request/[requestId].tsx
Comment (OustanDing) [reply pages/admin/request/[requestId].tsx:153]: Good catch, will remove this!

PR #48, opened 2021-08-06
[Improvement] Seperate daypicker component into its own module

EmilioMena:
## Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/Move-DayPicker-component-to-its-own-module-e3a11df0a5ae4d02a3fc8478102e0750)

## Implementation description
* Move `DayPicker` from permit-holders page into `@components/DayPicker.tsx` since it's used in other places

## Notes
* Didn't do too much in terms of abstracting the logic from DayPicker so that it can still be customized. Really only the styling is abstracted
* Simplified permit-holders page a bit as well

## Checklist


components/DayPicker.tsx
```diff
@@ -0,0 +1,39 @@
...
+        }
+      `}</style>
+    </Helmet>
+  );
+}
+
+/**
+ * DayPicker component from {@link http://react-day-picker.js.org/| react-day-picker} library with additional styling.
+ */
+export default function DayPicker(props: DayPickerProps) {
```
OustanDing, line 32: Given the purposes and use cases for the DayPicker of our team, I think we can abstract this a bit further. For instance, all our day picker components will have the same styling, and we just need to define the `modifiers` and the `onDayClick` props. We don't actually need to pass the `{ start: range.from || undefined, end: range.to || undefined }` value to the `DayPicker` component - we could just abstract this and have our `DayPicker` component take a `range` prop which is an object.

So our props for the `DayPicker` component could look something like this:

```typescript
type DayPickerProps {
  readonly numberOfMonths?: number; // # of months, default to 1
  readonly range: RangeModifier;
  readonly onDateChange: (day: <whatever the type is>) => void; // Handle date change
}
```

This way, we can abstract most of the functionality in the `DayPicker` component in this file. We can also even abstract our `range` state by writing our own custom hook `useDayPicker`. This will allow us to also hide away functions such as `dateRangeString()` below, so that we don't repeat that logic in every module in which we use the `DayPicker` component.

Would be happy to sync up with you @EmilioMena on this when I'm done my finals!

pages/admin/permit-holders.tsx
```diff
@@ -335,14 +304,11 @@ export default function PermitHolders() {
                 </MenuButton>
                 <MenuList>
                   <DayPicker
-                    className="Selectable"
-                    numberOfMonths={2}
                     selectedDays={[range.from || undefined, range]}
-                    modifiers={modifier}
-                    onDayClick={handleDayClick}
+                    modifiers={{ start: range.from || undefined, end: range.to || undefined }}
+                    onDayClick={day => setRange(DateUtils.addDayToRange(day, range))}
```
OustanDing, line 309: Please see comment above!

OustanDing (commented):
Left a comment - we can abstract a lot of the logic and achieve consistent usage for this component across our project. Let's sync up soon!

components/DayPicker.tsx
```diff
@@ -0,0 +1,39 @@
+import Helmet from 'react-helmet'; // Date picker inline styling for select range functionality
+import ReactDayPicker, { DayPickerProps } from 'react-day-picker';
+import 'react-day-picker/lib/style.css'; // Date picker styling
```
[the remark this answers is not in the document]
EmilioMena: yep the react-day-picker styling styles the calendar and the custom styles are for the select date range select functionality. I'll add that as a comment

EmilioMena:
@OustanDing I abstracted the dayPicker component out more and added the custom hook - let me know what you think!

components/DayPicker.tsx
EmilioMena: Good point! I tried address this in my last commit, lmk if i can add anything else

tools/hooks/useDayPicker.ts
```diff
@@ -0,0 +1,42 @@
...
+  dateRangeString: () => string;
+};
+
+/**
+ * Custom hook for managing DayPicker state
+ * @param {useDayPicker} initialDateRange - Initial date range object with from and to properties
+ * @returns dateRange state, addDayToDateRange function, dateRangeString function to display dates (YYYY-MM-DD - YYYY-MM-DD)
+ */
+export default function useDayPicker(initalDateRange: useDayPicker): useDayPickerReturnType {
+  const [dateRange, setDateRange] = useState<RangeModifier>(initalDateRange);
```
OustanDing, line 23:
```suggestion
export default function useDayPicker(initalDateRange?: InitialDateRange): UseDayPickerReturnType {
  const [dateRange, setDateRange] = useState<RangeModifier>(initalDateRange || { from: undefined, to: undefined });
```
We should make the initial date range optional and set the above as the default value. Following this change, we don't manually need to pass in the object with undefined `from` and `to` parameters as was done in the date range picker component.

tools/hooks/useDayPicker.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  function addDayToDateRange(day: Date) {
```
OustanDing, line 25: Could you please add a comment to this function? Its function is a little ambiguous

tools/hooks/useDayPicker.ts
```diff
@@ -0,0 +1,42 @@
...
+    setDateRange(DateUtils.addDayToRange(day, dateRange));
+  }
+
+  const dateRangeString = () => {
```
OustanDing, line 29: This can be a value, as it will be updated whenever the `dateRange` state gets updated. We can do `let dateRangeString` and conditionally set the value as below. Then, we don't need to call the function every time we need the `dateRangeString`

OustanDing (changes requested):
Some small things following our Slack discussion @EmilioMena

PR #54, opened 2021-08-28
[Misc] Migrate staging deployment to Netlify, fix build errors

OustanDing:
## Notion ticket link

[Ticket](https://www.notion.so/uwblueprintexecs/Migrate-staging-to-Netlify-Heroku-Postgres-7dde07559a044866bc7ee2ccff1ff6fd)

## Implementation description
* Move deployment from Railway to Netlify. Set up a new Heroku Postgres instance and linked it to the Netlify project
* I believe that due to recent upgrades to the `typescript` and `@types/node` packages, we were encountering a bunch of TypeScript-related build errors (especially around exception handling), cc @EmilioMena . Addressed those errors in this PR by checking that DB errors are instances of `Prisma.PrismaClientKnownRequestError` and introducing a helper function `getUniqueConstraintFailedField` for use when a DB operation fails a unique constraint.
* Added `target: 'experimental-serverless-trace'` to our Next.js config. This stops Next.js from inlining our dependencies and allows proper bundling during builds. I believe this is causing longer build times, however, so I will continue to look into any changes regarding this build issue. Please see https://github.com/netlify/netlify-plugin-nextjs/issues/527#issuecomment-879300081
* Moved some dependencies from dev dependencies to prod dependencies. These were needed in the build process. 
* Migrated GraphQL schemas from their `schema.graphql` files to respective `schema.ts` TypeScript files that export strings. These strings were then merged in `lib/graphql/type-defs.ts`. This was necessary because the `schema.graphql` files are no longer present during the build process.

## Notes
* We are migrating to Netlify because Railway's free tier is not sufficient for our needs
* Netlify also has PR deploys and an interactive feature review UI. I will keep these disabled for now to save on build minutes, but will activate them when it makes sense to do so.

## Checklist


prisma/index.ts
```diff
@@ -1,13 +1,12 @@
 import { PrismaClient } from '@prisma/client'; // Prisma ORM
 
-// add prisma to the NodeJS global type
-interface CustomNodeJsGlobal extends NodeJS.Global {
-  prisma: PrismaClient;
+// Add Prisma to the NodeJS global type
+declare global {
+  // var must be used for globally scoped variables
+  // eslint-disable-next-line no-var
+  var prisma: PrismaClient;
```
EmilioMena, line 7: curious why was this changes? Does it also prevent multiple instances of prisma client in development?

EmilioMena (approved):
Looks great 🚢 
Big thank you for doing all of this man. Really appreciate the effort you put into getting this done!

prisma/index.ts
OustanDing: Yeah good question. The most recent update to `@types/node` removed the `NodeJS.Global` type TypeScript gave us errors when extending it. Including link to SO post here: https://stackoverflow.com/a/68328402

OustanDing:
Thanks a lot for the quick review @EmilioMena ! Greatly appreciated :D

PR #58, opened 2021-09-14
[Fix] Fix incorrect dates displayed

j-bunk:
## Notion ticket link

[Permit Holders table showing incorrect date of birth](https://www.notion.so/uwblueprintexecs/Permit-Holders-table-showing-incorrect-date-of-birth-0e58daefb7da4e28a2470f27cd6fab41)
Ticket co-assignee @EmilioMena 

## Implementation description
* Dates were being converted to a different time zone when displayed which set the date back by one day ( e.g date of birth should be September 11th -> was shown as September 10th because of time zone conversion). To address this a util function `formatDate` was created that explicitly set the time zone to UTC.
* I swept through the codebase and refactored how we showed dates to use the `formatDate` function instead which makes most of the dates in the app have a MM/DD/YYYY format (1st pic)
* The dates on the Date received column in the requests table were changed from a MM/DD/YYYY to a verbose/written format to match the Figma designs (2nd pic)

## Design Review Pics + Notes
* Dates are now formatted in MM/DD/YYYY format as seen in the following picture:
image

* The date received column in the requests table is an exception where it's written more verbosely (matches Figma design):
image

* Worth noting exceptions to the MM/DD/YYYY date format are dates in date Input elements and dates in the date range picker which are in YYYY/MM/DD format
image
image

## Checklist

### DIFF
--- a/components/permit-holders/modals/EditUserInformationModal.tsx
+++ b/components/permit-holders/modals/EditUserInformationModal.tsx
@@ -20,6 +20,7 @@ import {
 import { useState, useEffect, ReactNode, SyntheticEvent } from 'react'; // React
 import { Gender, UpdateApplicantInput } from '@lib/graphql/types'; // Gender Enum
 import { ApplicantData } from '@tools/pages/admin/permit-holders/types'; // Applicant data type
+import { formatDate } from '@lib/utils/format'; // Date formatter util
 
 type EditUserInformationModalProps = {
   applicant: ApplicantData;
@@ -37,7 +38,7 @@ export default function EditUserInformationModal({
   // Personal information state
   const [firstName, setFirstName] = useState('');
   const [lastName, setLastName] = useState('');
-  const [dateOfBirth, setDateOfBirth] = useState(new Date().toISOString().substring(0, 10));
+  const [dateOfBirth, setDateOfBirth] = useState(formatDate(new Date(), true));
   const [gender, setGender] = useState<Gender | undefined>();
 
   // Contact information state
@@ -57,7 +58,7 @@ export default function EditUserInformationModal({
   useEffect(() => {
     setFirstName(applicant.firstName);
     setLastName(applicant.lastName);
-    setDateOfBirth(new Date(applicant.dateOfBirth).toISOString().substring(0, 10));
+    setDateOfBirth(formatDate(applicant.dateOfBirth, true));
     setGender(applicant.gender);
     setEmail(applicant.email || '');
     setPhoneNumber(applicant.phone);
--- a/components/permit-holders/modals/EditUserInformationModal.tsx
+++ b/components/permit-holders/modals/EditUserInformationModal.tsx
@@ -20,7 +20,6 @@ import {
 import { useState, useEffect, ReactNode, SyntheticEvent } from 'react'; // React
 import { Gender, UpdateApplicantInput } from '@lib/graphql/types'; // Gender Enum
 import { ApplicantData } from '@tools/pages/admin/permit-holders/types'; // Applicant data type
-import { formatDate } from '@lib/utils/format'; // Date formatter util
 
 type EditUserInformationModalProps = {
   applicant: ApplicantData;
@@ -38,7 +37,7 @@ export default function EditUserInformationModal({
   // Personal information state
   const [firstName, setFirstName] = useState('');
   const [lastName, setLastName] = useState('');
-  const [dateOfBirth, setDateOfBirth] = useState(formatDate(new Date()));
+  const [dateOfBirth, setDateOfBirth] = useState(new Date().toISOString().substring(0, 10));
   const [gender, setGender] = useState<Gender | undefined>();
 
   // Contact information state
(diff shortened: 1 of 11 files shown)


j-bunk:
I need to manually check the pages to make sure I didn't break anything. Will do that tomorrow (added a WIP label for now)

lib/utils/format.ts
```diff
@@ -24,3 +24,35 @@ export const formatPostalCode = (postalCode: string): string => {
...
+  });
+
+  // Removes first comma and writes am/pm in lowercase
+  const formattedDate =
+    localeDateString.replace(',', '').substring(0, localeDateString.length - 3) +
+    localeDateString[localeDateString.length - 2].toLowerCase() +
+    localeDateString[localeDateString.length - 1].toLowerCase();
+
+  return formattedDate;
+};
```
EmilioMena, line 58: What do you think about having both of these in the same function and adding a boolean parameter: `verbose` 🤔  
Imo having one function would make the code more concise and easier to read

EmilioMena:
I also think it would be nice to get a quick design review before merging!

lib/utils/format.ts
OustanDing: Having one function would allow us to keep all the date logic in one place. However, the date formatting logic is quite different in these two util functions. If we need to make changes in the future or maintain the logic, having them decoupled is better as it's slightly easier to isolate issues.

components/permit-holders/modals/EditUserInformationModal.tsx
```diff
@@ -58,7 +57,7 @@ export default function EditUserInformationModal({
   useEffect(() => {
     setFirstName(applicant.firstName);
     setLastName(applicant.lastName);
-    setDateOfBirth(formatDate(applicant.dateOfBirth));
+    setDateOfBirth(new Date(applicant.dateOfBirth).toISOString().substring(0, 10));
```
j-bunk: Reverted to how it was originally, using the formatting function displays the following before selecting a date
image

Instead of displaying the initial date of birth (before editing)
image

components/requests/modals/EditReasonForReplacementModal.tsx
```diff
@@ -37,7 +36,7 @@ export default function EditReasonForReplacementModal({
   const [reason, setReason] = useState('');
 
   //   Lost Information state
-  const [date, setDate] = useState(formatDate(new Date()));
+  const [date, setDate] = useState(new Date().toISOString().substring(0, 10));
```
j-bunk: See above comment

pages/verify-identity.tsx
```diff
@@ -48,7 +47,7 @@ export default function IdentityVerificationForm() {
 
   const [userId, setUserId] = useState('');
   const [phoneNumberSuffix, setPhoneNumberSuffix] = useState('');
-  const [dateOfBirth, setDateOfBirth] = useState(formatDate(new Date()));
+  const [dateOfBirth, setDateOfBirth] = useState(new Date().toISOString().substring(0, 10));
```
j-bunk: See above comment. The `formatDate` function displays yyyy-mm-dd instead of pre-filling with the current date which was the original behavior

components/permit-holders/modals/EditUserInformationModal.tsx
j-bunk: is now ` const [dateOfBirth, setDateOfBirth] = useState(formatDate(new Date(), true));`. Passing true into the `formatDate` function executes `new Date(applicant.dateOfBirth).toISOString().substring(0, 10)`

lib/utils/format.ts
```diff
@@ -24,3 +24,38 @@ export const formatPostalCode = (postalCode: string): string => {
...
+/**
+ * Format date to be in MM/DD/YYYY format and in UTC time zone to avoid the date being set back by a day
+ * @param {Date} date date to be formatted
+ * @param {boolean} dateInput Whether the date is being displayed in Input element of type date
+ * @returns {string} formatted date
+ */
+export const formatDate = (date: Date, dateInput = false): string => {
+  return dateInput
+    ? new Date(date).toISOString().substring(0, 10)
+    : new Date(date).toLocaleDateString('en-US', { timeZone: 'UTC' });
```
angeladietz, line 37: I'm guessing this probably doesn't make a difference but should we use 'en-ca' for the locale instead of 'en-us'?

lib/utils/format.ts
```diff
@@ -24,3 +24,38 @@ export const formatPostalCode = (postalCode: string): string => {
...
+};
+
+/**
+ * Format date to be in written in the following format: Sep 11 2021, 03:07 pm
+ * @param {Date} date date to be formatted
+ * @returns {string} formatted verbose date
+ */
+export const formatDateVerbose = (date: Date): string => {
+  const localeDateString = new Date(date).toLocaleDateString('en-US', {
```
angeladietz, line 46: Same thing here with the locale

angeladietz (approved):

lib/utils/format.ts
j-bunk: en-CA formats it with dashes (MM-DD-YYYY) while en-US formats the dates slashes MM/DD/YYYY. Based on feedback from design we want to use slashes so that's why I opted with en-US

lib/utils/format.ts
j-bunk: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString

Similar to above comment, but the en-US formats it differently than en-CA

PR #60, opened 2021-09-15
[Fix] Update physician on permit holder page when MSP is updated

angeladietz:
## Notion ticket link

[Resolve physician update failing in View Permit Holder page](https://www.notion.so/uwblueprintexecs/Resolve-physician-update-failing-in-View-Permit-Holder-page-3dab29ed46204525a7aa9849d5b60ed2)

## Implementation description
* When the Doctor's MSP number is updated through the edit doctor's information modal on the permit holder page, upset physician will create a new Physician in the DB. We then also call an `updateMedicalInformation` mutation to point the applicant's medical information to the new physician.

## Notes
* Currently the past physicians list from an applicant's medical history is not being updated properly with new physicians. The past physicians are being read from the list of applicant's past applications, so when a new physician is created in through this edit doctor modal, it will not appear in the list of past physicians as it's not associated with an application. This will be addressed in a later ticket.

## Checklist

### DIFF
--- a/pages/admin/permit-holder/[permitHolderId].tsx
+++ b/pages/admin/permit-holder/[permitHolderId].tsx
@@ -13,7 +13,7 @@ import MedicalHistoryCard from '@components/permit-holders/MedicalHistoryCard'; 
 import { GetPermitHolderRequest, GetPermitHolderResponse } from '@tools/pages/permit-holders/types';
 import { GET_PERMIT_HOLDER } from '@tools/pages/permit-holders/queries'; // Permit holder query
 import { useMutation, useQuery } from '@apollo/client'; // Apollo
-import { useState } from 'react'; // React
+import React, { useState } from 'react'; // React
 import {
   PermitData,
   MedicalHistoryEntry,
@@ -30,6 +30,11 @@ import {
   UpdateApplicantResponse,
   UPDATE_APPLICANT_MUTATION,
 } from '@tools/pages/admin/permit-holders/update-applicant'; // Update applicant types
+import {
+  UpdateMedicalInformationRequest,
+  UpdateMedicalInformationResponse,
+  UPDATE_MEDICAL_INFORMATION_MUTATION,
+} from '@tools/pages/admin/permit-holders/update-medical-information';
 
 type Props = {
   readonly permitHolderId: number;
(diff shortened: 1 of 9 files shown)


pages/admin/permit-holder/[permitHolderId].tsx
```diff
@@ -30,84 +30,106 @@ import {
...
+      physicianId = data.upsertPhysician.physicianId;
     },
     onError: error => {
       toast({
         status: 'error',
         description: error.message,
       });
+      setIsErrorOnUpdateDoctor(true);
     },
-    refetchQueries: ['GetPermitHolder'],
```
Comment (EmilioMena) [inline pages/admin/permit-holder/[permitHolderId].tsx:110]: why was this removed? Do we not have to refetch `GetPermitHolder` anymore?

pages/admin/permit-holder/[permitHolderId].tsx
```diff
@@ -30,84 +30,106 @@ import {
...
+  UpdateMedicalInformationResponse,
+  UPDATE_MEDICAL_INFORMATION_MUTATION,
+} from '@tools/pages/admin/permit-holders/update-medical-information';
 
 type Props = {
   readonly permitHolderId: number;
 };
 
+// Must be defined outside of PermitHolder function otherwise variable does not maintain value
+let physicianId: number;
```
Comment (EmilioMena) [inline pages/admin/permit-holder/[permitHolderId].tsx:44]: +1 with what Oustan said at standup, would be great to find a different way to get around this

EmilioMena (approved):
Great job y'all! 🚀  Left a few minor comments and questions. 
I might be missing some context on this PR but it LGTM from what I see

pages/admin/permit-holder/[permitHolderId].tsx
```diff
@@ -136,8 +158,33 @@ export default function PermitHolder({ permitHolderId }: Props) {
...
+      const oldDoctorMSP = data.applicant.medicalInformation.physician.mspNumber;
+      setIsErrorOnUpdateDoctor(false);
+
+      await submitEditedDoctorInformation({ variables: { input: { ...physicianData } } });
+
+      if (!isErrorOnUpdateDoctor && physicianData.mspNumber !== oldDoctorMSP) {
+        await submitUpdatedMedicalInformation({
+          variables: {
+            input: {
+              applicantId: +data.applicant.id,
```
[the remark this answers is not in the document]
Comment (angeladietz) [reply pages/admin/permit-holder/[permitHolderId].tsx:170]: It parses the value to be an int if it isn't already one (https://stackoverflow.com/questions/14667713/how-to-convert-a-string-to-number-in-typescript). Without it, we get a server error that the applicant id was wrapped in quotes.

pages/admin/permit-holder/[permitHolderId].tsx
Comment (angeladietz) [reply pages/admin/permit-holder/[permitHolderId].tsx:110]: In the case of the map number changing, we make two separate queries, and we would only want to refetch that query after the second one. Instead, we call `refetch()` at the end of `handleUpdateDoctorInformation` so that we only refetch once regardless of whether one or two queries were made.

OustanDing (approved):
LGTM!

PR #67, opened 2021-09-24
[Fix] Requests Table search by user ID not working

EmilioMena:
## Notion ticket link

[Requests Table search by user ID not working](https://www.notion.so/uwblueprintexecs/Requests-Table-search-by-user-ID-not-working-7e0fcd020c51498f987c9afc11864cb3)

rcd1

## Implementation description
* Changed the `applications` resolver to filter by the `rcdUserId` column instead of the `id` column of the `applicants` table
* Made variable names more clear to avoid confusion in the future

## Notes
* I might be missing some context but I'm not sure why the applications resolver was looking at the `applicants` table to begin with. Are there any implications of filtering by the `rcdUserId` field in the `applications` table instead?

## Checklist

### DIFF
--- a/lib/applicants/resolvers.ts
+++ b/lib/applicants/resolvers.ts
@@ -49,7 +49,7 @@ export const applicants: Resolver = async (_parent, { filter }, { prisma }) => {
 
     let expiryDateUpperBound,
       expiryDateLowerBound,
-      userIDSearch,
+      rcdUserIDSearch,
       nameFilters,
       firstSearch,
       middleSearch,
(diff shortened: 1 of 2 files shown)


OustanDing:
> I might be missing some context but I'm not sure why the applications resolver was looking at the applicants table to begin with. Are there any implications of filtering by the rcdUserId field in the applications table instead?

Good question. I'm guessing there was a misunderstanding on which ID to search by, though we could've done the search by `applicantId`. Anyway, we shouldn't need to look at the `applicant` relation.

OustanDing (approved):
LGTM!

PR #69, opened 2021-09-25
[Feature] Get Employee Endpoint

aaggarwal10:
## Notion ticket link

[Get Employee Endpoint](https://www.notion.so/uwblueprintexecs/ac7b1f636a214a0cbf45f6b76f899b54?v=5414ee5fc5974f149921e192d6eb29b8&p=98582f0df40b4e67866b6e9f456b6b81)

## Implementation description
Normal Query:
GetEmployee In GQPlayground
Null Case:
GetEmployee with Invalid ID

## Checklist

### DIFF
--- a/lib/employees/resolvers.ts
+++ b/lib/employees/resolvers.ts
@@ -11,7 +11,7 @@ import { DBErrorCode, getUniqueConstraintFailedFields } from '@lib/db/errors'; /
 export const employee: Resolver = async (_parent, args, { prisma }) => {
   const employee = await prisma.employee.findUnique({
     where: {
-      id: parseInt(_args.id),
+      id: parseInt(args.id),
     },
   });
   return employee;
--- a/lib/employees/resolvers.ts
+++ b/lib/employees/resolvers.ts
@@ -6,7 +6,7 @@ import { DBErrorCode, getUniqueConstraintFailedFields } from '@lib/db/errors'; /
 
 /**
  * Query for one employee in the internal-facing app given id
- * @returns employee row
+ * @returns employee if found, null otherwise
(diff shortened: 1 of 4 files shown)


lib/employees/resolvers.ts
```diff
@@ -4,6 +4,19 @@ import { Resolver } from '@lib/resolvers'; // Resolver type
 import { EmployeeAlreadyExistsError } from '@lib/employees/errors'; // Employee errors
 import { DBErrorCode, getUniqueConstraintFailedFields } from '@lib/db/errors'; // Database errors
 
+/**
+ * Query for one employee in the internal-facing app given id
+ * @returns employee row
```
OustanDing, line 9:
```suggestion
 * @returns employee if found, null otherwise
```

lib/employees/resolvers.ts
```diff
@@ -4,6 +4,19 @@ import { Resolver } from '@lib/resolvers'; // Resolver type
...
+ */
+export const employee: Resolver = async (_parent, _args, { prisma }) => {
```
OustanDing, line 11:
```suggestion
export const employee: Resolver = async (_parent, args, { prisma }) => {
```
Since we're using the `args` function parameter here, we don't want to prefix it with the underscore.

OustanDing (changes requested):
Resolvers look good! WE just need to add the respective types to the GraphQL schema. Let's sync up on this

OustanDing (approved):
LGTM!

OustanDing:
Just for clarification: some of the schemas for this endpoint were already built so they weren't included in this PR.

PR #75, opened 2021-10-01
[Feature] Added (Soft) Delete Employee Endpoint.

aaggarwal10:
## Notion ticket link

[Delete employee endpoint](calendar.google.com/calendar/u/0/r/week?eid=MnZodjRwZmVjbmJtZWdsM2tmbDhibXJkdTMgYW5pc2hhZ2dhcndhbEB1d2JsdWVwcmludC5vcmc&ctok=YW5pc2hhZ2dhcndhbEB1d2JsdWVwcmludC5vcmc&pli=1)

## Implementation description
* Added Resolver for checking condition of employee with given ID in database
* Updated active state if correct qualifications, otherwise threw error.
* Conditions include:
1. Employee does not exist in DB -> EmployeeNotFoundError
2. Employee already deleted from DB -> EmployeeAlreadyDeletedError

## Pictures
* Case: Deleting Non-existant User
Mutation
ResultantError
* Case: Normal Deletion
Mutation
Result
* Case: Double Deletion
Mutation
image

P.S. : I noticed that it says ID undefined in message, but that has been fixed due to a type. :D
i.e.
image

## Checklist

### DIFF
--- a/lib/employees/resolvers.ts
+++ b/lib/employees/resolvers.ts
@@ -1,7 +1,11 @@
 import { ApolloError } from 'apollo-server-errors'; // Apollo error
 import { Prisma } from '@prisma/client'; // Prisma client
 import { Resolver } from '@lib/resolvers'; // Resolver type
-import { EmployeeAlreadyExistsError, EmployeeNotFoundError } from '@lib/employees/errors'; // Employee errors
+import {
+  EmployeeAlreadyDeletedError,
+  EmployeeAlreadyExistsError,
+  EmployeeNotFoundError,
+} from '@lib/employees/errors'; // Employee errors
 import { DBErrorCode, getUniqueConstraintFailedFields } from '@lib/db/errors'; // Database errors
 import { SortOrder } from '@tools/types'; // Sorting Type
 
@@ -118,4 +122,42 @@ export const updateEmployee: Resolver = async (_, args, { prisma }) => {
   return {
     ok: true,
   };
+};
+
+/**
+ * Delete an employee
+ * @returns status of operation (ok)
+ */
+export const deleteEmployee: Resolver = async (_, args, { prisma }) => {
+  const id = parseInt(args.id);
+  const employee = await prisma.employee.findUnique({
+    where: {
+      id,
+    },
+    select: {
+      active: true,
+    },
+  });
+
+  if (employee === null) {
+    throw new EmployeeNotFoundError(`Employee with ID ${id} not found`);
+  } else if (!employee.active) {
+    throw new EmployeeAlreadyDeletedError(`Employee with ID ${id} already deleted`);
+  }
+
+  let updatedEmployee;
+  try {
+    updatedEmployee = await prisma.employee.update({
+      where: { id },
+      data: { active: false },
+    });
+  } catch (err) {
+    if (err instanceof Prisma.PrismaClientKnownRequestError) {
+      if (err.code === DBErrorCode.RecordNotFound) {
+        throw new EmployeeNotFoundError(`Employee with ID ${id} not found`);
+      }
+    }
+  }
+  if (!updatedEmployee) throw new ApolloError('Employee was unable to be deleted');
+  return { ok: true };
 };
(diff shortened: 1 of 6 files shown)


lib/employees/resolvers.ts
```diff
@@ -119,3 +123,29 @@ export const updateEmployee: Resolver = async (_, args, { prisma }) => {
...
+/**
+ * Delete an employee
+ * @returns status of operation (ok)
+ */
+export const deleteEmployee: Resolver = async (_, args, { prisma }) => {
+  const employee = await prisma.employee.findUnique({
+    where: {
+      id: parseInt(args.id),
+    },
+  });
```
OustanDing, line 136: We can optimize this query by selecting only the fields we need (`active`). Please see https://www.prisma.io/docs/concepts/components/prisma-client/select-fields#select-specific-fields

OustanDing: Also can we add some blank lines to this resolver for readability please?

OustanDing (changes requested):
So far so good! We still need to add the types to the appropriate schemas a regenerate TypeScript types though

lib/employees/resolvers.ts
```diff
@@ -119,3 +123,35 @@ export const updateEmployee: Resolver = async (_, args, { prisma }) => {
...
+    throw new EmployeeNotFoundError(`Employee with ID ${parseInt(id)} not found`);
+  } else if (!employee.active) {
+    throw new EmployeeAlreadyDeletedError(`Employee with ID ${parseInt(id)} already deleted`);
+  } else {
+    const employee = await prisma.employee.update({
+      where: { id: parseInt(id) },
+      data: { active: false },
+    });
+
+    if (employee.active) throw new ApolloError('Employee was unable to be deleted');
```
OustanDing, line 153: Pretty sure that `employee` is not in sync with the database so I don't think this will work?

It would be better to take everything in the `else` block out of it and put it in a `try-catch` block instead with the `ApolloError` being thrown in the `catch` block.

lib/graphql/schema.ts
```diff
@@ -16,6 +16,7 @@ export default `
     updateApplicant(input: UpdateApplicantInput!): UpdateApplicantResult!
     createEmployee(input: CreateEmployeeInput!): CreateEmployeeResult!
     updateEmployee(input: UpdateEmployeeInput!): UpdateEmployeeResult!
+    deleteEmployee(input: DeleteEmployeeInput!): DeleteEmployeeResult!
```
OustanDing, line 19: Since there is only one input parameter (ID) we can just pass it in directly, no need for putting it in an object:
```suggestion
    deleteEmployee(id: ID!): DeleteEmployeeResult!
```
and the `DeleteEmployeeInput` type should be removed from `lib/employees/schema.ts`

OustanDing (changes requested):

lib/employees/resolvers.ts
```diff
@@ -119,3 +123,41 @@ export const updateEmployee: Resolver = async (_, args, { prisma }) => {
...
+  const employee = await prisma.employee.findUnique({
+    where: {
+      id: id,
+    },
+    select: {
+      active: true,
+    },
+  });
+
+  if (employee == null) {
```
OustanDing, line 142:
```suggestion
  if (employee === null) {
```
This will give us type comparison as well.

lib/employees/resolvers.ts
```diff
@@ -119,3 +123,41 @@ export const updateEmployee: Resolver = async (_, args, { prisma }) => {
...
+
+/**
+ * Delete an employee
+ * @returns status of operation (ok)
+ */
+export const deleteEmployee: Resolver = async (_, args, { prisma }) => {
+  const id = parseInt(args.id);
+  const employee = await prisma.employee.findUnique({
+    where: {
+      id: id,
```
OustanDing, line 135:
```suggestion
      id,
```

lib/employees/resolvers.ts
```diff
@@ -119,3 +123,41 @@ export const updateEmployee: Resolver = async (_, args, { prisma }) => {
...
+  if (employee == null) {
+    throw new EmployeeNotFoundError(`Employee with ID ${id} not found`);
+  } else if (!employee.active) {
+    throw new EmployeeAlreadyDeletedError(`Employee with ID ${id} already deleted`);
+  }
+
+  let updatedEmployee;
+  try {
+    updatedEmployee = await prisma.employee.update({
+      where: { id: id },
```
OustanDing, line 151:
```suggestion
      where: { id },
```

OustanDing (changes requested):

OustanDing (approved):
LGTM!

PR #84, opened 2021-10-09
[Feature] Admin Management Table API Hookup

chrischan325:
## Notion ticket link

[Admin Management Table API Hookup - Delete admin](https://www.notion.so/uwblueprintexecs/Admin-Management-Table-API-Hookup-Delete-admin-05121d7aff88479691d2155c1c5dcc84)

## Implementation description
* Created the gql delete admin mutation 
* hooked the api up to the delete admin modal
* added `notifyOnNetworkStatusChange` to the getEmployees useQuery hook to update the admin-management table upon deletion

## Notes
* Made a slight change to the deleteEmployee resolver to make sure the ID is recieved through args and it is parsed to be an integer
* Will add comments to added code later 
* Lines 142 and 143 in `admin-management.tsx` are not fixed because that would be part of the update api hookup ticket. Not sure if I should fix it in this PR, would like a second opinion pls!

## Checklist

### DIFF
--- a/lib/employees/resolvers.ts
+++ b/lib/employees/resolvers.ts
@@ -149,7 +149,7 @@ export const updateEmployee: Resolver = async (_, args, { prisma }) => {
  * @returns status of operation (ok)
  */
 export const deleteEmployee: Resolver = async (_, args, { prisma }) => {
-  const id = parseInt(args.input);
+  const id = parseInt(args.id);
 
   const employee = await prisma.employee.findUnique({
     where: {
--- a/lib/employees/resolvers.ts
+++ b/lib/employees/resolvers.ts
@@ -149,7 +149,7 @@ export const updateEmployee: Resolver = async (_, args, { prisma }) => {
  * @returns status of operation (ok)
  */
 export const deleteEmployee: Resolver = async (_, args, { prisma }) => {
-  const id = parseInt(args.input.id);
+  const id = parseInt(args.input);
 
   const employee = await prisma.employee.findUnique({
     where: {
@@ -180,5 +180,5 @@ export const deleteEmployee: Resolver = async (_, args, { prisma }) => {
     }
   }
   if (!updatedEmployee) throw new ApolloError('Employee was unable to be deleted');
-  return { ok: true };
+  return { ok: true, employee: updatedEmployee };
 };
--- a/lib/employees/resolvers.ts
+++ b/lib/employees/resolvers.ts
@@ -149,7 +149,8 @@ export const updateEmployee: Resolver = async (_, args, { prisma }) => {
  * @returns status of operation (ok)
  */
 export const deleteEmployee: Resolver = async (_, args, { prisma }) => {
-  const id = parseInt(args.id);
+  const id = parseInt(args.input.id);
+
   const employee = await prisma.employee.findUnique({
     where: {
       id,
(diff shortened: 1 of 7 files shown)


lib/employees/schema.ts
```diff
@@ -36,6 +36,10 @@ export default `
     employee: Employee!
   }
 
+  input DeleteEmployeeInput {
+    id: ID!
+  }
```
OustanDing, line 41: Why do we need to nest the `id` argument in an object? When deleting an employee, we would only require one parameter which is the ID, so an object shouldn't be needed

lib/employees/resolvers.ts
```diff
@@ -149,7 +149,8 @@ export const updateEmployee: Resolver = async (_, args, { prisma }) => {
  * @returns status of operation (ok)
  */
 export const deleteEmployee: Resolver = async (_, args, { prisma }) => {
-  const id = parseInt(args.id);
+  const id = parseInt(args.input.id);
+
```
OustanDing, line 153: Sorry this wasn't included in the original ticket, but for the update and create endpoints, we're returning the updated/created employee record from the API so that we can grab the name of the employee that was affected. We want to do the same for this endpoint, ie. return the deleted employee, to enforce data integrity (getting the data from the DB instead of the frontend). Please see https://github.com/uwblueprint/richmond-centre-for-disability/pull/85 for an example

pages/admin/admin-management.tsx
```diff
@@ -303,7 +346,7 @@ export default function AdminManagement() {
           onClose={onCloseConfirmDeleteModal}
           user={userToDelete}
           // TODO: Replace onDelete handler during API hookup
```
OustanDing, line 348: Remove this TODO comment

pages/admin/admin-management.tsx
```diff
@@ -262,8 +268,43 @@ export default function AdminManagement() {
...
 
+  // Delete Employee Hook
+  const [deleteEmployee] = useMutation<DeleteEmployeeResponse, DeleteEmployeeRequest>(
+    DELETE_EMPLOYEE_MUTATION,
+    {
+      onCompleted: data => {
+        if (data.deleteEmployee.ok) {
+          toast({
+            status: 'success',
+            description: `${userToDelete?.name} has been deleted`,
```
OustanDing, line 287: No longer use `userToDelete`. Use the `employee` from the `deleteEmployee` mutation result

pages/admin/admin-management.tsx
```diff
@@ -262,8 +268,43 @@ export default function AdminManagement() {
...
+  );
+
+  // Handle delete employee action
+  const handleDelete = async () => {
+    if (userToDelete) {
+      await deleteEmployee({
+        variables: {
+          input: userToDelete?.id,
+        },
+      });
```
OustanDing, line 301: Same as above

lib/employees/resolvers.ts
```diff
@@ -149,7 +149,8 @@ export const updateEmployee: Resolver = async (_, args, { prisma }) => {
  * @returns status of operation (ok)
  */
 export const deleteEmployee: Resolver = async (_, args, { prisma }) => {
-  const id = parseInt(args.id);
+  const id = parseInt(args.input);
```
OustanDing, line 152:
```suggestion
  const id = parseInt(args.id);
```
Need this change after reverting input variable name to `id`

OustanDing (changes requested):

PR #102, opened 2021-11-10
[Feature] Generate Report Button & Modal for Requests and Permit Holder Page

chrischan325:
## Notion ticket link

[Build Generate Report Button and Report Modal Frontend for Requests and Permit Holder Table pages](https://www.notion.so/uwblueprintexecs/Build-Generate-Report-Button-and-Report-Modal-Frontend-for-Requests-and-Permit-Holder-Table-pages-15b7d474e5e243c08859331f88ab13e9)

## Implementation description
* Used enums to switch the modal view from selecting columns to the export as csv view
* Since state for the checkboxes aren't included in this ticket, the select all checkbox doesn't work and the text on the second view of the modal are all default text.

## Notes
* FOR DESIGNERS: since we switched the date picker component to the html picker, I can't put start date/end date in the box, let me know if the form heading is fine for the date fields. Also, using the html picker means that the formatting of the date is different, which is why it is shown as YYYY-MM-DD.

generate-report-demo1

generate-report-demo2

Screen Shot 2021-11-10 at 8 43 09 PM

Screen Shot 2021-11-10 at 8 53 18 PM

Screen Shot 2021-11-10 at 8 43 50 PM

Screen Shot 2021-11-10 at 8 44 09 PM

## Checklist


components/admin/permit-holders/modals/GenerateReportModal.tsx
```diff
@@ -0,0 +1,193 @@
...
+
+type GenerateReportProps = {
+  readonly isOpen: boolean;
+  readonly onClose: () => void;
+};
+
+enum GenerateReportModalEnum {
+  SELECT_COLUMNS = 1,
+  EXPORT = 2,
+}
```
OustanDing, line 30: Put this in a file in `tools/`

components/admin/permit-holders/modals/GenerateReportModal.tsx
```diff
@@ -0,0 +1,193 @@
...
+} from '@chakra-ui/react'; // Chakra UI
+import { DownloadIcon } from '@chakra-ui/icons';
+import { useState } from 'react'; // React
+
+type GenerateReportProps = {
+  readonly isOpen: boolean;
+  readonly onClose: () => void;
+};
+
+enum GenerateReportModalEnum {
```
OustanDing, line 27:
```suggestion
enum GenerateReportStep {
```

components/admin/permit-holders/modals/GenerateReportModal.tsx
```diff
@@ -0,0 +1,193 @@
...
+export default function GenerateReportModal(props: GenerateReportProps) {
+  const { isOpen, onClose } = props;
+  const [startDate, setStartDate] = useState('');
+  const [endDate, setEndDate] = useState('');
+  const [step, setStep] = useState(GenerateReportModalEnum.SELECT_COLUMNS);
+
+  if (step == 1) {
+    return (
+      <>
+        <Modal isOpen={isOpen} onClose={onClose} isCentered size={'xl'}>
```
OustanDing, line 41: We don't want each step to render a new `Modal` instance. Instead, what we can do is return a `Modal` and make a function `renderStep` which takes in a `step` parameter (of the enum type), and the `renderStep` function returns the content for the modal step.

components/admin/permit-holders/modals/GenerateReportModal.tsx
```diff
@@ -0,0 +1,193 @@
...
+  EXPORT = 2,
+}
+
+export default function GenerateReportModal(props: GenerateReportProps) {
+  const { isOpen, onClose } = props;
+  const [startDate, setStartDate] = useState('');
+  const [endDate, setEndDate] = useState('');
+  const [step, setStep] = useState(GenerateReportModalEnum.SELECT_COLUMNS);
+
+  if (step == 1) {
```
OustanDing, line 38:
```suggestion
  if (step == GenerateReportModalEnum.SELECT_COLUMNS) {
```

components/admin/permit-holders/modals/GenerateReportModal.tsx
OustanDing: Also make the enum keys PascalCase

OustanDing (changes requested):
In addition to the comments I left, we only need one of the two modals - let's put it in `components/admin/reports/GenerateReportModal.tsx`. When API hookup comes around, we'll add props such as `onChange`, `onSubmit`, `options` to make it usable with both requests and permit holders.

OustanDing (approved):
LGTM!

PR #109, opened 2021-11-25
[Feature] Build CSV generation API for requests

j-bunk:
## Notion ticket link

[Reports - Build CSV generation API for requests](https://www.notion.so/uwblueprintexecs/ac7b1f636a214a0cbf45f6b76f899b54?v=5414ee5fc5974f149921e192d6eb29b8)

## Implementation description

- Created a GraphQL query endpoint `generateApplicantsReport` that takes in an input object `input` with the following fields (all required):
    - `startDate` 
  - `endDate` 
  - `columns` - array of `ApplicationsReportColumn` enum
 - Created `ApplicationsReportColumn` enum in `lib/graphql/schema.ts`  with appropriate enum values
- Used Prisma to query the `applications` table and filter the records by `createdAt` timestamp
- Used `csv-writer` to subsequently generate a CSV file

The generated file is saved in the file system for now.

## Demo
Query:
image

Generated csv:
image

[file.csv](https://github.com/uwblueprint/richmond-centre-for-disability/files/7638140/file.csv)

## Notes
* See comments

## Checklist


lib/applications/resolvers.ts
```diff
@@ -547,3 +549,95 @@ export const createReplacementApplication: Resolver = async (_, args, { prisma }
...
+    where: {
+      createdAt: {
+        gte: new Date(startDate),
+        // TODO; Figure out why lte isn't working
+        // lte: new Date(endDate),
+      },
+    },
+    select: {
+      // TODO?: Replace with function/object in constants
+      rcdUserId: columns.includes(ApplicationsReportColumn.UserId),
```
j-bunk, line 570: I think making columns an object instead of an array could be better. We'll avoid inefficient array lookup and it would make the code cleaner:
e.g
```ts
 select: {
      rcdUserId: columns.userId,
```

lib/applications/resolvers.ts
```diff
@@ -547,3 +549,95 @@ export const createReplacementApplication: Resolver = async (_, args, { prisma }
...
+        ? ` ${application.middleName} ${application.lastName}`
+        : ` ${application.lastName}`);
+    application.totalAmount = (application.processingFee || 0) + (application?.donationAmount || 0);
+    application.rcdPermitId = application.permits?.rcdPermitId;
+    return application;
+  });
+
+  const csvHeaders = [];
+
+  if (columns.includes(ApplicationsReportColumn.UserId)) {
```
j-bunk, line 605: https://github.com/uwblueprint/richmond-centre-for-disability/pull/109/files#r757935443

lib/applications/resolvers.ts
```diff
@@ -547,3 +549,93 @@ export const createReplacementApplication: Resolver = async (_, args, { prisma }
...
 };
+
+export const generateApplicantsReport: Resolver = async (_, args, { prisma }) => {
+  const {
+    input: { startDate, endDate, columns },
+  } = args;
+
+  const columnsSet = new Set(columns);
+
+  const applications: GenerateApplicantsReportApplications[] = await prisma.application.findMany({
```
j-bunk, line 560: TypeScript was throwing errors when I tried adding `totalAmount`, `applicantName` and `rcdPermitId` properties to the `applications` object (https://github.com/uwblueprint/richmond-centre-for-disability/blob/jb/csv-generation-api-requests/lib/applications/resolvers.ts#L589-L599) so I had to create a new type (`GenerateApplicantsReportApplications`) with those properties and explicitly set `applications` to that type

lib/applications/resolvers.ts
```diff
@@ -547,3 +549,93 @@ export const createReplacementApplication: Resolver = async (_, args, { prisma }
...
+        columnsSet.has(ApplicationsReportColumn.FeeAmount) ||
+        columnsSet.has(ApplicationsReportColumn.TotalAmount),
+      donationAmount:
+        columnsSet.has(ApplicationsReportColumn.DonationAmount) ||
+        columnsSet.has(ApplicationsReportColumn.TotalAmount),
+      permits: {
+        select: {
+          rcdPermitId: true,
+        },
+      },
```
j-bunk, line 585: A prisma `select` statement needs at least one truthy value so the nested select has to be true here or else the query fails

lib/applications/types.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * applications object type used in csv generatiion
+ */
+export type GenerateApplicantsReportApplications = {
```
j-bunk, line 4: lol does anyone have better name suggestions

lib/applications/resolvers.ts
j-bunk: Converted the array in the resolver based on oustan's suggestion

lib/graphql/schema.ts
```diff
@@ -9,6 +9,7 @@ export default `
     applications(filter: ApplicationsFilter): QueryApplicationsResult
     application(id: ID!): Application
     permits: [Permit!]
+    generateApplicantsReport(input: GenerateApplicantsReportInput!): GenerateApplicantsReportResult!
```
j-bunk, line 12: The notion ticket said that generateApplicantsReport should be a mutation but I think it makes more sense as a query since no data was modified. lmk if it should be changed to a mutation instead

lib/applications/resolvers.ts
```diff
@@ -611,3 +613,93 @@ export const createReplacementApplication: Resolver = async (_, args, { prisma }
...
```
j-bunk, line 649: The notion ticket suggests to 

> Create an object in lib/applications/constants.ts that maps the ApplicationsReportColumn enums to strings that Prisma would consume when selecting columns in a DB query

I didn't implement that but I could extract the highlighted lines into a function if you think that makes the code cleaner

lib/graphql/schema.ts
```diff
@@ -119,4 +120,16 @@ export default `
     PERMANENT
     TEMPORARY
   }
+
+  enum ApplicationsReportColumn {
+    USER_ID
+    APPLICANT_NAME
+    APPLICANT_DOB
```
OustanDing, line 127:
```suggestion
    APPLICANT_DATE_OF_BIRTH
```

lib/graphql/schema.ts
OustanDing: That makes sense. Good catch.

OustanDing:
```suggestion
    generateApplicantsReport(input: GenerateApplicantsReportInput!): GenerateApplicantsReportResult
```
Needs to be nullable if authorization fails

lib/applications/schema.ts
```diff
@@ -350,4 +350,14 @@ export default `
...
     ok: Boolean!
   }
+
+  input GenerateApplicantsReportInput {
+    startDate: Date!
+    endDate: Date!
+    columns: [ApplicationsReportColumn!]!
+  }
+
+  type GenerateApplicantsReportResult {
```
OustanDing, line 360: Can we add a comment here saying that in the future this should also return a link to an S3 file?

lib/applications/resolvers.ts
```diff
@@ -611,3 +613,93 @@ export const createReplacementApplication: Resolver = async (_, args, { prisma }
...
+    },
+  });
+
+  // Adds totalAmount, applicantName and rcdPermitId properties to applications object
+  applications.map(application => {
```
OustanDing, line 654: Here we can make a new variable that is a result of this map so that we don't need to modify the original data. This requires more space but is easier to maintain and eliminates TypeScript errors

lib/applications/resolvers.ts
```diff
@@ -611,3 +613,93 @@ export const createReplacementApplication: Resolver = async (_, args, { prisma }
...
+    application.applicantName =
+      application.firstName +
+      (application.middleName
+        ? ` ${application.middleName} ${application.lastName}`
+        : ` ${application.lastName}`);
```
OustanDing, line 659:
```suggestion
    application.applicantName = `${application.firstName} ${application.middleName || ''} ${application.lastName}`
```

lib/applications/resolvers.ts
OustanDing: Your current approach works fine as well! We can optimize later

lib/applications/types.ts
OustanDing: See comment above in `lib/applications/resolvers.ts`, I don't think we'll need this anymore

j-bunk: Removed!

lib/applications/schema.ts
j-bunk: Done!

lib/applications/resolvers.ts
j-bunk: Pretty minor issue but I noticed with this approach we end up with two spaces between firstName and lastName if middleName is null.

lib/applications/resolvers.ts
j-bunk: Good suggestion! Done!

lib/applications/resolvers.ts
OustanDing: Okay, sounds good

OustanDing (approved):
LGTM! Great work!

PR #119, opened 2021-12-12
[Feature] Build CSV generation API for permit holders

j-bunk:
## Notion ticket link

[Reports - Build CSV generation API for permit holders](https://www.notion.so/uwblueprintexecs/Reports-Build-CSV-generation-API-for-permit-holders-96e83e2175714b4a87e69c7301d1d0af)

## Implementation description

- Created a GraphQL query endpoint `generatePermitHoldersReport` that takes in an input object `input` with the following fields (all required):
    - `startDate` 
  - `endDate` 
  - `columns` - array of `PermitHoldersReportColumn` enum values
 - Created `PermitHoldersReportColumn` enum in `lib/graphql/schema.ts`  with appropriate enum values
- Used Prisma to query the `applicants` table and filter the records by `createdAt` timestamp of applicants' applications
- Used `csv-writer` to subsequently generate a CSV file

The generated file is saved in the file system for now.

### Demo:
GraphQL Query Results:
image

Generated CSV (UPDATED):
[file-permit-holders.csv](https://github.com/uwblueprint/richmond-centre-for-disability/files/7765398/file-permit-holders.csv)

Screenshot of generated csv (UPDATED):
image

## Notes
* Could you take a look at the csv and lmk if you want anything to be formatted differently
* Left some comments with questions

## Checklist

### DIFF
--- a/lib/applicants/resolvers.ts
+++ b/lib/applicants/resolvers.ts
@@ -10,10 +10,11 @@ import {
 import { getUniqueConstraintFailedFields, DBErrorCode } from '@lib/db/errors'; // Database errors
 import { MspNumberDoesNotExistError } from '@lib/physicians/errors'; // Physician errors
 import { getActivePermit } from '@lib/applicants/utils'; // Applicant utils
-import { formatPhoneNumber, formatPostalCode } from '@lib/utils/format'; // Formatting utils
-import { PermitStatus, VerifyIdentityFailureReason } from '@lib/types'; // GraphQL types
+import { formatDate, formatPhoneNumber, formatPostalCode } from '@lib/utils/format'; // Formatting utils
+import { PermitHoldersReportColumn, PermitStatus, VerifyIdentityFailureReason } from '@lib/types'; // GraphQL types
 import { DateUtils } from 'react-day-picker'; // Date utils
 import { SortOrder } from '@tools/types'; // Sorting Type
+import { createObjectCsvWriter } from 'csv-writer';
 
 /**
  * Query and filter RCD applicants from the internal facing app.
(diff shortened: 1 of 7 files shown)


lib/applications/resolvers.ts
```diff
@@ -713,3 +717,129 @@ export const generateApplicationsReport: Resolver = async (_, args, { prisma })
...
+      guardianPostalCode: columnsSet.has(PermitHoldersReportColumn.GuardianPoaAddress),
+      guardianCity: columnsSet.has(PermitHoldersReportColumn.GuardianPoaAddress),
+      guardianProvince: columnsSet.has(PermitHoldersReportColumn.GuardianPoaAddress),
+      permit: {
+        select: {
+          rcdPermitId: true,
+          // NOTETOSELF: Is user status whether the permit is active,
+          // or is it the ApplicationStatus (taken from ApplicationProcessing),
+          // or is it the ApplicantStatus (taken from Applicant)
+          active: columnsSet.has(PermitHoldersReportColumn.UserStatus),
```
j-bunk, line 763: @OustanDing I wasn't sure what User Status maps to in the db. Would you happen to know?
image

lib/applications/resolvers.ts
```diff
@@ -713,3 +717,131 @@ export const generateApplicationsReport: Resolver = async (_, args, { prisma })
...
+      guardianAddressLine1: columnsSet.has(PermitHoldersReportColumn.GuardianPoaAddress),
+      guardianAddressLine2: columnsSet.has(PermitHoldersReportColumn.GuardianPoaAddress),
+      guardianPostalCode: columnsSet.has(PermitHoldersReportColumn.GuardianPoaAddress),
+      guardianCity: columnsSet.has(PermitHoldersReportColumn.GuardianPoaAddress),
+      guardianProvince: columnsSet.has(PermitHoldersReportColumn.GuardianPoaAddress),
+      permit: {
+        select: {
+          rcdPermitId: true,
+          active: columnsSet.has(PermitHoldersReportColumn.UserStatus),
+        },
```
j-bunk, line 761: I'm fetching the `active` field from the permit table and using that as the User Status. Is this correct or should be fetching a different field from a different table
@OustanDing

lib/applications/resolvers.ts
```diff
@@ -713,3 +717,131 @@ export const generateApplicationsReport: Resolver = async (_, args, { prisma })
...
+        (application.guardianMiddleName
+          ? ` ${application.guardianMiddleName} ${application.guardianLastName}`
+          : ` ${application.guardianLastName}`),
+      guardianPOAAdress: `${
+        application.guardianAddressLine1 ? `${application.guardianAddressLine1},` : ''
+      } ${application.guardianAddressLine2 ? `${application.guardianAddressLine2},` : ''} ${
+        application.guardianCity ? `${application.guardianCity},` : ''
+      } ${application.guardianProvince ? `${application.guardianProvince},` : ''} ${
+        application.guardianPostalCode ?? ''
+      }`,
```
j-bunk, line 794: This is the value that's written to the csv for guardian/poa address.
This separates the address values with commas (and all the address values are nullable so I used a ternary operator). This doesn't look great but works lol, lmk if you have any suggestions to improve this.

lib/graphql/types.ts
```diff
@@ -438,9 +438,9 @@ export type CreateReplacementApplicationInput = {
   postalCode: Scalars['String'];
   /** Replacement Information */
   reason: ReasonForReplacement;
-  lostTimestamp: Scalars['Date'];
-  lostLocation: Scalars['String'];
-  description: Scalars['String'];
+  lostTimestamp?: Maybe<Scalars['Date']>;
+  lostLocation?: Maybe<Scalars['String']>;
+  description?: Maybe<Scalars['String']>;
```
j-bunk, line 443: Not related to my changes

lib/applications/resolvers.ts
OustanDing: There is a `status` column in the `applicants` table that should correspond to the user status (active/inactive)

lib/applications/resolvers.ts
```diff
@@ -713,3 +717,131 @@ export const generateApplicationsReport: Resolver = async (_, args, { prisma })
...
+      applicantName:
+        application.firstName +
+        (application.middleName
+          ? ` ${application.middleName} ${application.lastName}`
+          : ` ${application.lastName}`),
+      rcdPermitId: application.permit?.rcdPermitId,
+      homeAddress: `${application.addressLine1}, 
+      ${application.addressLine2 ? `${application.addressLine2},` : ''} ${application.city}, ${
+        application.province
+      }, ${application.postalCode}`,
```
OustanDing, line 781:
```suggestion
      } ${application.postalCode}`,
```
Usually, we don't write a comma between province and postal code

lib/applications/resolvers.ts
OustanDing: This works haha, perhaps we can make it a bit readable using my above comment, but no big deal if we can't

lib/applications/resolvers.ts
```diff
@@ -713,3 +717,131 @@ export const generateApplicationsReport: Resolver = async (_, args, { prisma })
...
+  }
+  if (columnsSet.has(PermitHoldersReportColumn.GuardianPoaAddress)) {
+    csvHeaders.push({ id: 'guardianPOAAdress', title: 'Guardian/POA Address' });
+  }
+  if (columnsSet.has(PermitHoldersReportColumn.RecentAppNumber)) {
+    csvHeaders.push({ id: 'rcdPermitId', title: 'Recent APP Number' });
+  }
+  if (columnsSet.has(PermitHoldersReportColumn.RecentAppType)) {
+    csvHeaders.push({ id: 'permitType', title: 'Recent APP Type' });
+  }
```
OustanDing, line 832: For most recent APP type, it should be on permits, not applications. Our DB currently doesn't support this, but I will add a ticket to the backlog to address this. For now, could you please use the `permit_type` column on the most recent application and add a TODO comment to update this? Thanks :)

OustanDing (commented):
A few comments

lib/applications/resolvers.ts
j-bunk: Addressed!

lib/applications/resolvers.ts
```diff
@@ -713,3 +717,131 @@ export const generateApplicationsReport: Resolver = async (_, args, { prisma })
...
 };
+
+export const generatePermitHoldersReport: Resolver = async (_, args, { prisma }) => {
+  const {
+    input: { startDate, endDate, columns },
+  } = args;
+
+  const columnsSet = new Set(columns);
+
+  const applications = await prisma.application.findMany({
```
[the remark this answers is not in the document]
j-bunk: Addressed!

lib/applications/resolvers.ts
```diff
@@ -713,3 +717,131 @@ export const generateApplicationsReport: Resolver = async (_, args, { prisma })
...
+          rcdPermitId: true,
+          active: columnsSet.has(PermitHoldersReportColumn.UserStatus),
+        },
+      },
+      permitType: columnsSet.has(PermitHoldersReportColumn.RecentAppType),
+    },
+  });
+
+  // Formats the date fields and adds totalAmount, applicantName and rcdPermitId properties to allow for csv writing
+  const csvApplications = applications.map(application => {
```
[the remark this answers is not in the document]
j-bunk: Addressed!

> We should either 1) modify the former step to just select all columns (so that latter step can reliably format all the required fields), or 2) build the objects using a bunch of if conditions so account for some fields not being selected in the former step. While approach 1 gives better performance, approach 2 is easier to understand and maintain. For now, to prioritize simplicity, let's just select all the columns and all necessary nested columns, and keep the current logic for the mapping.


Just to clarify,  is this supposed to be the opposite (approach 2 gives better performance, and approach 1 is easier to understand)?

lib/applications/resolvers.ts
```diff
@@ -713,3 +717,131 @@ export const generateApplicationsReport: Resolver = async (_, args, { prisma })
...
+    return {
+      ...application,
+      dateOfBirth: formatDate(application.dateOfBirth),
+      applicantName:
+        application.firstName +
+        (application.middleName
+          ? ` ${application.middleName} ${application.lastName}`
+          : ` ${application.lastName}`),
```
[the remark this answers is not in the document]
j-bunk: Addressed!

lib/applications/resolvers.ts
j-bunk: Addressed!

lib/applications/resolvers.ts
j-bunk: Addressed with the following:

image

lib/applications/resolvers.ts
j-bunk: Addressed with the following code
image

lib/applications/resolvers.ts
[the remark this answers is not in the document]
j-bunk: Hmm thought about this a bit more there doesn't seem to be an issue with the original implementation, calling the mapping function on fields that we never selected doesn't break anything and returns mostly null values (which won't be written to the csv since they were never selected). 

So with the original implementation we have both better performance (since we only select/fetch the fields we want) and also have simplicity since we don't write if statements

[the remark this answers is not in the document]
OustanDing: > Just to clarify, is this supposed to be the opposite (approach 2 gives better performance, and approach 1 is easier to understand)?
Sorry for the late response, you're correct in this statement! Sorry for the confusion.

My original issue was that we were using `columnsSet.has` for every `select` which is very verbose. But I think your current implementation has a good balance between readability and performance, so let's stick with that. Thanks for pointing this out :)

lib/applicants/resolvers.ts
```diff
@@ -390,3 +391,161 @@ export const verifyIdentity: Resolver = async (_, args, { prisma }) => {
...
+      dateOfBirth: formatDate(applicant.dateOfBirth),
+      applicantName: `${applicant.firstName}${
+        applicant.middleName ? ` ${applicant.middleName}` : ''
+      } ${applicant.lastName}`,
+      rcdPermitId: applicant.permits[0].rcdPermitId,
+      permitType: applicant.applications[0].permitType,
+      homeAddress: `${applicant.addressLine1},${
+        applicant.addressLine2 ? ` ${applicant.addressLine2},` : ''
+      } ${applicant.city}, ${applicant.province} ${applicant.postalCode}`,
+      guardianRelationship: applicant.guardian?.relationship,
```
OustanDing, line 486: Just want to double check, will this print `undefined`?

OustanDing (approved):

lib/applicants/resolvers.ts
OustanDing: Just tested, it will print empty string.

OustanDing:
Permit holders filter should be based on active permit expiry date, not application creation date

PR #112, opened 2021-12-04
[Feature] PhysicianAssessment and GuardianInformation new app forms

angeladietz:
## Notion ticket link

[New APP Requests Page - PhysicianAssessment and GuardianInformation forms frontend](https://www.notion.so/uwblueprintexecs/New-APP-Requests-Page-PhysicianAssessment-and-GuardianInformation-forms-frontend-c3d4be78ef7245a0a0d2368e7e61a842)

## Implementation description
* Created 2 new form components in `components/requests/forms/new-applications/PhysicianAssessmentForm.tsx` and `components/requests/forms/new-applications/GuardianInformationForm.tsx`

Screen Shot 2021-12-05 at 3 01 03 PM

Screen Shot 2021-12-04 at 11 16 54 AM

## Notes
* 

## Checklist
- [x] My PR name is descriptive, is in imperative tense and starts with one of the following: `[Feature]`,`[Improvement]` or `[Fix]`,
- [x] I have run the appropriate linter(s)
- [x] I have requested a review from the RCD team on GitHub, or specific people who are associated with this ticket

PR #121, opened 2021-12-17
[Feature] Add email updates checkbox to PermitHolderInformationForm

angeladietz:
## Notion ticket link

[Add email updates checkbox to PermitHolderInformationForm](https://www.notion.so/uwblueprintexecs/Add-email-updates-checkbox-to-PermitHolderInformationForm-38eb5e3aab72488099da694be2d9e147)

## Implementation description
* Add `recieveEmailUpdates` field to `Applications`
* Create 2 types of props for `PermitHolderInformationForm`, `ReplacementPermitHolderInformationFormProps` and `NewAndRenewalPermitHolderInformationFormProps` and dynamically access the props depending on the type specified.
* Add checkbox to `PermitHolderInformationForm` when request type is not `replacement`

## Notes

Screen Shot 2021-12-20 at 4 11 02 PM

## Checklist


lib/graphql/resolvers.ts
```diff
@@ -83,8 +83,8 @@ const resolvers = {
     createPhysician: authorize(createPhysician, [Role.Secretary]),
     upsertPhysician: authorize(upsertPhysician, [Role.Secretary]),
     createApplication: authorize(createApplication, [Role.Secretary]),
-    createRenewalApplication,
-    createReplacementApplication,
+    createRenewalApplication: authorize(createRenewalApplication, [Role.Secretary]),
+    createReplacementApplication: authorize(createReplacementApplication, [Role.Secretary]),
```
OustanDing, line 87: Good catch here. `createRenewalApplication` shouldn't be wrapped in `authorize` though, because permit holders can apply for a renewal themselves

lib/applications/schema.ts
```diff
@@ -111,6 +112,7 @@ export default `
     rcdUserId: Int
     isRenewal: Boolean!
     permitType: PermitType!
+    receiveEmailUpdates: Boolean
```
OustanDing, line 115:
```suggestion
    receiveEmailUpdates: Boolean!
```
For new applications, this boolean should be non-nullable

lib/applications/schema.ts
```diff
@@ -321,6 +324,7 @@ export default `
     addressLine2: String
     city: String!
     postalCode: String!
+    receiveEmailUpdates: Boolean
```
OustanDing, line 327:
```suggestion
```
We should remove this boolean from `CreateReplacementApplicationInput` since the checkbox isn't present for replacement applications

components/admin/requests/forms/PermitHolderInformationForm.tsx
```diff
@@ -36,26 +67,40 @@ export default function PermitHolderInformationForm({
...
+                  props.onChange({
+                    ...props.permitHolderInformation,
+                    firstName: event.target.value,
+                  });
+                } else {
+                  props.onChange({
+                    ...props.permitHolderInformation,
+                    firstName: event.target.value,
+                  });
+                }
```
OustanDing, line 82: Are these conditionals necessary?

Also to reduce the amount of duplicated code (doing object destructuring for every form field), we can make a higher-order function as discussed in this thread: https://github.com/uwblueprint/richmond-centre-for-disability/pull/112#discussion_r773466014

components/admin/requests/modals/EditPermitHolderInformationModal.tsx
```diff
@@ -62,6 +62,7 @@ export default function EditPermitHolderInformationModal({
             </ModalHeader>
             <ModalBody paddingY="20px" paddingX="4px">
               <PermitHolderInformationForm
+                type="replacement"
```
OustanDing, line 65: Shouldn't we pass down the request type as a prop through this Modal component, rather than hard-coding it to `replacement`?

components/admin/requests/forms/PermitHolderInformationForm.tsx
angeladietz: Yep the conditions are necessary, even though the if and else cases are the same

components/admin/requests/modals/EditPermitHolderInformationModal.tsx
angeladietz: Yep good point, fixed now!

components/admin/requests/forms/PermitHolderInformationForm.tsx
OustanDing: Hm interesting, we will look into this in the future then

components/admin/requests/modals/EditPermitHolderInformationModal.tsx
```diff
@@ -61,10 +114,20 @@ export default function EditPermitHolderInformationModal({
...
+                  onChange={handleReplacementPermitHolderInformationChange}
+                />
+              )}
+              {permitHolderTypeAndInformation.type !== 'replacement' && (
+                <PermitHolderInformationForm
+                  type={permitHolderTypeAndInformation.type}
+                  permitHolderInformation={permitHolderTypeAndInformation.permitHolderInformation}
+                  onChange={handleNewAndRenewalPermitHolderInformationChange}
+                />
+              )}
```
OustanDing, line 130:
```suggestion
              {permitHolderTypeAndInformation.type === 'replacement' ? (
                <PermitHolderInformationForm
                  type={permitHolderTypeAndInformation.type}
                  permitHolderInformation={permitHolderTypeAndInformation.permitHolderInformation}
                  onChange={handleReplacementPermitHolderInformationChange}
                />
              ) : (
                <PermitHolderInformationForm
                  type={permitHolderTypeAndInformation.type}
                  permitHolderInformation={permitHolderTypeAndInformation.permitHolderInformation}
                  onChange={handleNewAndRenewalPermitHolderInformationChange}
                />
              )}
```

OustanDing (commented):
A few items

components/admin/requests/PersonalInformationCard.tsx
```diff
@@ -41,9 +42,18 @@ export default function PersonalInformationCard(props: Props) {
     </Link>
   );
 
+  /* eslint-disable @typescript-eslint/no-unused-vars */
```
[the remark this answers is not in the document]
angeladietz: I did this to remove those two fields from the `applicant` object because they were being passed to the `EditPermitHolderInformationModal` which was causing errors. The alternative that I see is manually specifying the full object to be passed into the `permitHolderInformation` on line 53 just below. Would you prefer this?

lib/applications/schema.ts
```diff
@@ -261,7 +264,7 @@ export default `
     usesAccessibleConvertedVan: Boolean!
     requiresWiderParkingSpace: Boolean!
     rcdUserId: Int
-
+    receiveEmailUpdates: Boolean
```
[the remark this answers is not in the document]
angeladietz: Right now this will break the non-admin renewal flow since theres no option for specifying this field yet. I'll still make this change but it will be hard coded to `false` (the default value) for now to be addressed later, similar to how `usesAccessibleConvertedVan` and `requiresWiderParkingSpace` are currently hard coded.

components/admin/requests/PersonalInformationCard.tsx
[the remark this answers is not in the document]
OustanDing: Okay, I see. That makes sense. I'd suggest the following for clarity:
```suggestion
  // Remove unwanted fields
  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```

lib/applications/schema.ts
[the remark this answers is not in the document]
OustanDing: That's a good workaround, thanks.

components/admin/requests/PersonalInformationCard.tsx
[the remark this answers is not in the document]
OustanDing: Addressed

OustanDing (approved):

PR #129, opened 2022-01-07
[Improvement] Misc DB schema changes

OustanDing:
## Notion ticket link

[Ticket](https://www.notion.so/uwblueprintexecs/Misc-DB-Changes-0832433795c148c1b0e16884d7c4cddb)

## Implementation description
Please see Notion ticket above and spreadsheet linked in Notion ticket

Summary:
- Updated DB schema with better decoupling, fixed affected APIs and frontend components
  - Improved and cleaned up types for frontend components and states
- Add strong TS checking for GraphQL resolvers, which is also usable on frontend
- Updated seeding scripts and grouped together similar models (eg. applicant + medical information + physician) to reduce number of seeding files and improve visibility
- Fixes and improvements to frontend components

## Notes
* Breaking changes
* Will help us streamline our progress in Term 3 so that we don't have to worry about big DB schema changes and frequent DB resets throughout the term

## Checklist


prisma/schema.sql
```diff
@@ -35,265 +48,371 @@ CREATE TYPE PhysicianStatus as ENUM('ACTIVE', 'INACTIVE');
...
-  app_holepunched     BOOLEAN NOT NULL DEFAULT false,
+  id INTEGER,
+  status ApplicationStatus NOT NULL DEFAULT 'PENDING',
+  rejected_reason VARCHAR(255),
+  app_number INTEGER UNIQUE,
+  app_number_employee_id INTEGER,
+  app_number_updated_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  app_holepunched BOOLEAN NOT NULL DEFAULT false,
+  app_number_employee_id INTEGER,
+  app_number_updated_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
angeladietz, line 218:
```suggestion
  app_holepunched_employee_id INTEGER,
  app_holepunched_updated_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
```

prisma/schema.sql
```diff
@@ -35,265 +48,371 @@ CREATE TYPE PhysicianStatus as ENUM('ACTIVE', 'INACTIVE');
...
   FOREIGN KEY(medical_information_id) REFERENCES medical_information(id)
 );
 
 -- Create application processing table
 CREATE TABLE application_processing (
-  id                  SERIAL PRIMARY KEY NOT NULL,
-  status              ApplicationStatus NOT NULL DEFAULT 'PENDING',
-  app_number          INTEGER UNIQUE,
-  app_holepunched     BOOLEAN NOT NULL DEFAULT false,
+  id INTEGER,
```
angeladietz, line 210: Why is this an `INTEGER` instead of `SERIAL`? (same with medical information, guardian, etc.)

prisma/schema.sql
```diff
@@ -35,265 +48,371 @@ CREATE TYPE PhysicianStatus as ENUM('ACTIVE', 'INACTIVE');
...
-  uses_accessible_converted_van  BOOLEAN DEFAULT FALSE,
-  requires_wider_parking_space   BOOLEAN DEFAULT FALSE,
-  application_id                 INTEGER UNIQUE NOT NULL,
-  created_at                     TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
-  updated_at                     TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  application_id INTEGER,
+  receive_email_updates BOOLEAN NOT NULL,
+
+  -- Physician information
+  physician_first_name VARCHAR(255) NOT NULL,
```
angeladietz, line 333: Nice separation of each of the application types! It looks good as is and will work. There's still a good amount of overlap between `renewals` and `new_applications`, so we could consider creating an additional table for these common fields (mostly just the physician info, receive email updates, and the van/parking space fields). I'm cool with it either way

prisma/schema.sql
```diff
@@ -35,265 +48,371 @@ CREATE TYPE PhysicianStatus as ENUM('ACTIVE', 'INACTIVE');
...
+  id INTEGER,
+  first_name VARCHAR(255) NOT NULL,
+  middle_name VARCHAR(255),
+  last_name VARCHAR(255) NOT NULL,
+  phone VARCHAR(50) NOT NULL,
+  address_line_1 VARCHAR(255) NOT NULL,
+  address_line_2 VARCHAR(255),
+  city VARCHAR(255) NOT NULL,
+  province Province NOT NULL DEFAULT 'BC',
+  country VARCHAR(255) NOT NULL DEFAULT 'Canada',
```
angeladietz, line 149: A concern about including the `country` field (across all tables): although it seems like a valid scenario for people to be outside of Canada, then the `province` field wouldn't cover the alternate country and we don't allow people to enter a non-canadian province equivalent, and the 6 character requirement on postal code would also be invalid (american zip codes are 5 digits i think)

angeladietz (commented):
Looks good! Left a few comments/questions

prisma/schema.sql
OustanDing: Good catch. My original intention was to make the `id`s the same value as the applicant's `id`, but that would run into issues when creating an applicant. Changed relevant instances to `SERIAL PRIMARY KEY`

prisma/schema.sql
OustanDing: That's a good point. While that reduces repetition, I think it makes the code harder to write and maintain. For instance, when creating a renewal application, we would have to use Prisma nested create, and it would be harder to debug/access an application record as we would need to access 3+ tables. Let's keep it this way for now

prisma/schema.sql
OustanDing: Great catch! Currently, we are assuming that all applicants, physicians, and guardians have addresses in BC, Canada - there is no UI for editing these fields, only for viewing. I agree though, we should add some flexibility to support the possibility of international addresses in the future. I will change the type of `postal_code` to `VARCHAR(255)`. Province can remain as an enum - it is easier to add support for more enum values than to change the column data type

prisma/schema.sql
```diff
@@ -35,269 +48,366 @@ CREATE TYPE PhysicianStatus as ENUM('ACTIVE', 'INACTIVE');
...
-  id                SERIAL PRIMARY KEY NOT NULL,
-  rcd_permit_id     INTEGER NOT NULL UNIQUE,
-  expiry_date       DATE NOT NULL,
-  receipt_id        INTEGER,
-  active            BOOLEAN NOT NULL DEFAULT true,
-  application_id    INTEGER NOT NULL UNIQUE,
-  applicant_id      INTEGER NOT NULL,
-  created_at        TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
-  updated_at        TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  rcd_permit_id INTEGER,
```
angeladietz, line 287: Should this one also be `SERIAL` or are rcd permit ids assignment differently?

OustanDing: `rcd_permit_id` is manually assigned by RCD, so when we create a permit from a completed application, we copy the assigned permit number and use that as the primary key

angeladietz (approved):
looks great!

jeffreyzhang2001 (approved):

PR #133, opened 2022-01-22
[Fix] Fix applicants endpoint filter

OustanDing:
## Notion ticket link

[Ticket](https://www.notion.so/uwblueprintexecs/Fix-applicants-API-filter-5f59658a83a14a6f9bc907f29de6d205)

## Implementation description
* `applicants` endpoint was returning no applicants when an empty filter was specified since when an empty filter was specified, only applicants with at least one permit were selected
* Conditionally built filter based on which filters were specified (expiry date, permit status)
* Also fixes a small bug in the permit holders table - if the most recent permit is `null`, "N/A" will now be rendered in the table cell instead of the entire page breaking

## Notes
* There is still a small bug that deals with timezones and date filtering, which we will further investigate during the date auditing process

## Checklist

### DIFF
--- a/lib/applicants/resolvers.ts
+++ b/lib/applicants/resolvers.ts
@@ -62,16 +62,12 @@ export const applicants: Resolver<
   let where = {};
 
   if (filter) {
-    const {
-      permitStatus = undefined,
-      userStatus = undefined,
-      expiryDateRangeFrom = undefined,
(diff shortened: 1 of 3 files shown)


angeladietz (approved):
Looks good, thanks for the fix!!

PR #136, opened 2022-01-24
[Feature] Accounting Report Generation API

bonnie-chin:
## Notion ticket link

[Build Accountant Reports Generation API](https://www.notion.so/uwblueprintexecs/Build-Accountant-Reports-Generation-API-698a110be5a045989838e72cf5792701)

[Desired CSV](https://docs.google.com/document/d/1zeNH_U_-bN5NsABBIq_0WMFrAMy4ED-ffbCUBNJVWcY/edit#heading=h.ydup5da146j8) - given by design

## Implementation description

Normal Query: 
image

Generated CSV: 
image

## Notes
- Second opinion: _is there a more optimal way of generating the leftmost column headers?_
- To-do: modify API to return link from AWS
- Note: file.csv is stored in temp for now

## Checklist

### DIFF
--- a/lib/graphql/resolvers.ts
+++ b/lib/graphql/resolvers.ts
@@ -122,7 +122,7 @@ const resolvers = {
     // Reports
     generateApplicationsReport: authorize(generateApplicationsReport, ['SECRETARY']),
     generatePermitHoldersReport: authorize(generatePermitHoldersReport, ['SECRETARY']),
-    generateAccountantReport: authorize(generateAccountantReport, [`SECRETARY`]),
+    generateAccountantReport: authorize(generateAccountantReport, [`ACCOUNTING`]),
   },
   Mutation: {
     // Applicants
--- a/lib/graphql/resolvers.ts
+++ b/lib/graphql/resolvers.ts
@@ -59,7 +59,11 @@ import {
   applicationProcessingResolver,
 } from '@lib/applications/field-resolvers'; // Application field resolvers
 import { medicalInformationPhysicianResolver } from '@lib/medical-information/field-resolvers';
-import { generatePermitHoldersReport, generateApplicationsReport } from '@lib/reports/resolvers';
+import {
+  generatePermitHoldersReport,
+  generateApplicationsReport,
+  generateAccountantReport,
+} from '@lib/reports/resolvers';
 import { permitApplicationResolver } from '@lib/permits/field-resolvers';
 
 /**
@@ -118,6 +122,7 @@ const resolvers = {
     // Reports
     generateApplicationsReport: authorize(generateApplicationsReport, ['SECRETARY']),
     generatePermitHoldersReport: authorize(generatePermitHoldersReport, ['SECRETARY']),
+    generateAccountantReport: authorize(generateAccountantReport, [`SECRETARY`]),
   },
   Mutation: {
     // Applicants
--- a/lib/reports/resolvers.ts
+++ b/lib/reports/resolvers.ts
@@ -301,7 +301,7 @@ export const generateAccountantReport: Resolver<
   }
   csvAccountantReportRows.push({
     rowName: 'Total',
-    countIssued: totalAggregate._count.paymentMethod,
+    countIssued: totalAggregate._count.paymentMethod || 0,
     processingFee: totalAggregate._sum.processingFee || 0,
     donationAmount: totalAggregate._sum.donationAmount || 0,
     totalAmount: Prisma.Decimal.add(
--- a/lib/reports/resolvers.ts
+++ b/lib/reports/resolvers.ts
@@ -5,9 +5,11 @@ import {
   GeneratePermitHoldersReportResult,
   QueryGenerateApplicationsReportArgs,
   GenerateApplicationsReportResult,
+  QueryGenerateAccountantReportArgs,
+  GenerateAccountantReportResult,
 } from '@lib/graphql/types';
 import { SortOrder } from '@tools/types';
-import { formatAddress, formatDate, formatFullName } from '@lib/utils/format'; // Formatting utils
+import { formatAddress, formatDate, formatFullName, formatPaymentType } from '@lib/utils/format'; // Formatting utils
 import { APPLICATIONS_COLUMNS, PERMIT_HOLDERS_COLUMNS } from '@tools/admin/reports';
 
 /**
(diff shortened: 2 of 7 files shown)


lib/graphql/resolvers.ts
```diff
@@ -118,6 +122,7 @@ const resolvers = {
     // Reports
     generateApplicationsReport: authorize(generateApplicationsReport, ['SECRETARY']),
     generatePermitHoldersReport: authorize(generatePermitHoldersReport, ['SECRETARY']),
+    generateAccountantReport: authorize(generateAccountantReport, [`SECRETARY`]),
```
OustanDing, line 125:
```suggestion
    generateAccountantReport: authorize(generateAccountantReport, [`ACCOUNTING`]),
```
This feature is for accounting team only

lib/utils/format.ts
```diff
@@ -105,3 +106,24 @@ ${country || ''}
 ${postalCode}
   `;
 };
+
+export const formatPaymentType = (paymentMethod: string) => {
```
OustanDing, line 110: Please add a JSDoc comment to this function, similar to how it's done for the other formatting functions in this file.

Also, please add a return type:
```suggestion
export const formatPaymentType = (paymentMethod: string): string => {
```

lib/utils/format.ts
```diff
@@ -105,3 +106,24 @@ ${country || ''}
...
+    case PaymentType.ETRANSFER: {
+      paymentMethod = 'E-Transfer';
+      break;
+    }
+    default: {
+      paymentMethod = paymentMethod[0] + paymentMethod.substring(1).toLowerCase();
+      break;
+    }
+  }
+  return paymentMethod + ending;
```
OustanDing, line 128: While most of the payment methods should just be converted from all capital to capitalized (eg. CASH => Cash), the more readable and easy to maintain/change solution is to just use an object and map the enums to their strings. Put such an object within this function:
```typescript
const paymentTypeToString: Record<PaymentType, string> = {
  'MASTERCARD': 'MasterCard (Office)',
  'VISA': 'Visa (Office)',
  ...rest
}
```

OustanDing: Also, please don't import `PaymentType` from `@prisma/client` - import it from `@lib/graphql/types` instead. Our enums are now string unions (see `lib/graphql/schema.ts` to provide better compatibility with select/radio fields and to give us type inference when using Prisma (both Prisma and GraphQL will use the same string values now).

When working with the `PaymentType` enum, if you specify the type when declaring the object (see above comment in this thread at `Record<PaymentType, string`), you will get type suggestions when writing the code.

lib/reports/resolvers.ts
```diff
@@ -244,3 +246,80 @@ export const generateApplicationsReport: Resolver<
...
+      },
+    },
+    _sum: {
+      processingFee: true,
+      donationAmount: true,
+    },
+    _count: {
+      paymentMethod: true,
+    },
+  });
```
OustanDing, line 277: Great!

lib/reports/resolvers.ts
```diff
@@ -244,3 +246,80 @@ export const generateApplicationsReport: Resolver<
...
+
+  const csvAccountantReportRows = [];
+  for (const key in paymentMethodGroups) {
+    const paymentMethodGroup = paymentMethodGroups[key];
```
OustanDing, line 291:
```suggestion
  for (const paymentMethodGroup of paymentMethodGroups) {
```
For arrays, use `of`. `in` is for objects

lib/reports/resolvers.ts
```diff
@@ -244,3 +246,80 @@ export const generateApplicationsReport: Resolver<
...
+    csvAccountantReportRows.push({
+      rowName: formatPaymentType(paymentMethodGroup.paymentMethod),
+      countIssued: paymentMethodGroup._count.paymentMethod,
+      processingFee: paymentMethodGroup._sum.processingFee,
+      donationAmount: paymentMethodGroup._sum.donationAmount,
```
OustanDing, line 296:
```suggestion
      processingFee: paymentMethodGroup._sum.processingFee || 0,
      donationAmount: paymentMethodGroup._sum.donationAmount || 0,
```
Since these are numeric, print "0" instead of nothing

lib/reports/resolvers.ts
```diff
@@ -244,3 +246,80 @@ export const generateApplicationsReport: Resolver<
...
+      totalAmount: paymentMethodGroup._sum.donationAmount?.add(
+        paymentMethodGroup._sum.processingFee || 0
+      ),
```
OustanDing, line 299:
```suggestion
      totalAmount: Prisma.Decimal.add(
        paymentMethodGroup._sum.donationAmount || 0, 
        paymentMethodGroup._sum.processingFee || 0,
      )
```
Make sure to add `import { Prisma } from '@prisma/client'` at the top of the file to have access to `Prisma.Decimal`. We are using `Prisma.Decimal` for better precision when adding monetary values (eg. with normal computer arithmetic we get 0.1 + 0.2 != 0.3). Since either donation amount or processing fee could be `null`, we should use the `Prisma.Decimal` method `add` in the way I suggested above.

In the current code, if `donationAmount` is `null`, `totalAmount` will automatically be `null` since the `.add` won't get evaluated.

See docs: https://mikemcl.github.io/decimal.js/#Dadd

lib/reports/resolvers.ts
```diff
@@ -244,3 +246,80 @@ export const generateApplicationsReport: Resolver<
...
+    });
+  }
+  csvAccountantReportRows.push({
+    rowName: 'Total',
+    countIssued: totalAggregate._count.paymentMethod,
+    processingFee: totalAggregate._sum.processingFee,
+    donationAmount: totalAggregate._sum.donationAmount,
+    totalAmount: totalAggregate._sum.donationAmount?.add(totalAggregate._sum.processingFee || 0),
```
OustanDing, line 307: Same as above comment

lib/reports/resolvers.ts
OustanDing: This might not seem like an issue now, but if somehow Prisma breaks or we receive no donations in some timespan, we don't want to have the report show that processing fee was also 0

lib/reports/resolvers.ts
```diff
@@ -244,3 +246,80 @@ export const generateApplicationsReport: Resolver<
...
+      totalAmount: paymentMethodGroup._sum.donationAmount?.add(
+        paymentMethodGroup._sum.processingFee || 0
+      ),
+    });
+  }
+  csvAccountantReportRows.push({
+    rowName: 'Total',
+    countIssued: totalAggregate._count.paymentMethod,
+    processingFee: totalAggregate._sum.processingFee,
+    donationAmount: totalAggregate._sum.donationAmount,
```
OustanDing, line 306:
```suggestion
    processingFee: totalAggregate._sum.processingFee || 0,
    donationAmount: totalAggregate._sum.donationAmount || 0,
```

lib/reports/resolvers.ts
```diff
@@ -244,3 +246,80 @@ export const generateApplicationsReport: Resolver<
...
+    totalAmount: totalAggregate._sum.donationAmount?.add(totalAggregate._sum.processingFee || 0),
+  });
+
+  const csvHeaders = [];
+  csvHeaders.push({ id: 'rowName', title: '' });
+  csvHeaders.push({ id: 'countIssued', title: 'Issued #' });
+  csvHeaders.push({ id: 'processingFee', title: 'Fees' });
+  csvHeaders.push({ id: 'donationAmount', title: 'Donation' });
+  csvHeaders.push({ id: 'totalAmount', title: 'Total' });
```
OustanDing, line 315: Can just declare `csvHeaders` with these members initially:
```typescript
const csvHeaders = [
  { id: 'rowName', title: '' },
  { id: 'countIssued', title: 'Issued #' },
 ...rest
]
```

OustanDing (changes requested):
Looks pretty good so far! Left some suggestions to improve maintainability and error avoidance

lib/utils/format.ts
OustanDing: Still missing JSDoc. Please see other formatting functions in this file

lib/reports/resolvers.ts
```diff
@@ -287,32 +287,36 @@ export const generateAccountantReport: Resolver<
...
+      donationAmount: paymentMethodGroup._sum.donationAmount || 0,
+      totalAmount: Prisma.Decimal.add(
+        paymentMethodGroup._sum.donationAmount || 0,
         paymentMethodGroup._sum.processingFee || 0
       ),
     });
   }
   csvAccountantReportRows.push({
     rowName: 'Total',
     countIssued: totalAggregate._count.paymentMethod,
```
OustanDing, line 304:
```suggestion
    countIssued: totalAggregate._count.paymentMethod || 0,
```

OustanDing (changes requested):
Need to add JSDoc comment to formatting function and make a small fix

lib/reports/resolvers.ts
aaggarwal10: Count can be null?

OustanDing (approved):

PR #139, opened 2022-02-05
[Feature] Create file upload form field

angeladietz:
## Notion ticket link

[Create file upload form field (frontend)](https://www.notion.so/uwblueprintexecs/Create-file-upload-form-field-frontend-340e597a3e8845e891fb56887da1d56e)

## Implementation description
* Created new component `FileUpload` and added it to the `GuardianInformationForm`

https://user-images.githubusercontent.com/35577524/152659143-ed3db1c3-5cd8-4de6-b75a-2ef3943bf8eb.mov

image

image

## Notes
* When there is a file uploaded and you click the button to replace it, if you click cancel the file will be removed. Not sure if this is the correct behaviour.

## Checklist


components/FileUploadField.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { Button } from '@chakra-ui/react';
+import { SyntheticEvent, useRef } from 'react';
+
+// File upload button props
+type FileUploadProps = {
+  currentFile: File | null; // currently uploaded file
+  onUploadFile: (selectedFile: File) => void; // handle file upload
+};
+
+//TODO: rename this better??
```
OustanDing, line 11: Let's call this `PoaFormUploadField` and put it in `components/admin/requests/guardian-information/PoaFormUploadField.tsx`

components/FileUploadField.tsx
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * File upload button component allowing users to upload one file
```
OustanDing, line 13:
```suggestion
 * POA form upload component allowing users to upload one file
```

components/FileUploadField.tsx
```diff
@@ -0,0 +1,50 @@
...
+ * @param props - props
+ * @returns File upload button component allowing users to upload one file
```
OustanDing, line 15: Can remove these since they don't add much value

components/FileUploadField.tsx
```diff
@@ -0,0 +1,50 @@
...
+ */
+export default function FileUpload(props: FileUploadProps) {
+  const hiddenFileInput = useRef<HTMLInputElement>(null);
+
+  const handleFileUpload = () => {
```
OustanDing, line 20: I would rename this to something like `handleClick` as it's more descriptive of the event being handled

components/FileUploadField.tsx
```diff
@@ -0,0 +1,50 @@
...
+    if (hiddenFileInput.current) {
+      hiddenFileInput.current.click();
+    }
+  };
+
+  const handleChange = (event: SyntheticEvent) => {
+    const target = event.target as HTMLInputElement;
+    const fileUploaded = (target.files as FileList)[0];
+    props.onUploadFile(fileUploaded ?? null);
+  };
```
OustanDing, line 30:
```suggestion
  const handleChange = (event: SyntheticEvent<HTMLInputElement>) => {
    const target = event.target;
    const fileUploaded = (target.files)[0];
    props.onUploadFile(fileUploaded ?? null);
  };
```
I believe we can use type generics here to avoid type casting

components/admin/requests/guardian-information/Form.tsx
```diff
@@ -153,15 +156,31 @@ export default function GuardianInformationForm({
             <Text as="h3" textStyle="heading" paddingBottom="20px">
               {'Upload POA File'}
             </Text>
-            {/* TODO: Implement the file upload functionality using the fileList prop */}
+            {/* TODO: explicitly prevent non pdfs or files > 5Mb? */}
```
OustanDing, line 159: We should be able to add this restriction within the file upload component. Please see https://stackoverflow.com/a/49490014. If it's a quick fix, it would be great to just add it in in this PR.

We can also restrict the file extension within the file upload component: https://stackoverflow.com/questions/4328947/limit-file-format-when-using-input-type-file

components/admin/requests/guardian-information/Form.tsx
```diff
@@ -153,15 +156,31 @@ export default function GuardianInformationForm({
...
-              {'Click to add attachement'}
-            </Button>
+            {file && (
+              <>
+                <Text as="h4" textStyle="button-semibold" mt="24px">
+                  Current File
+                </Text>
+                <Box mt="8px">
+                  <Link
+                    href={URL.createObjectURL(file)}
```
OustanDing, line 172: Just wondering - does this actually create a clickable link to open the file? What does this line do?

angeladietz: Yeah, it makes a temporary url so when you click the link it will download the file. 
https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL

components/FileUploadField.tsx
OustanDing: Actually I just realized that we're getting some linting errors if we remove the casting for some reason. Will leave it as is for now

OustanDing:
I moved the entire field title and description to be within the file upload component, as those should also be included in the field itself (so that they are reused). Also, added error states (please see updated screenshots)

OustanDing (approved):

PR #144, opened 2022-02-24
[Feature] Update confirmation email content

j-bunk:
## Notion ticket link

[Write HTML and CSS for confirmation email + SMS content based on copy](https://www.notion.so/uwblueprintexecs/Write-HTML-and-CSS-for-confirmation-email-SMS-content-based-on-copy-1ade14a070094a0aa9b7f22d729b7867)

## Implementation description
* Updated email content

Minor details:
* Changed the sender email to use an env variable called `CONFIRMATION_EMAIL_FROM`
* The application is also queried to fetch the receiver's first name and then use it in the confirmation email

## Email screenshots for design review:
I sort of eyeballed the alignment. Lmk if the alignment looks wonky. I tried a few other alignments, lmk if you prefer any to the current implementation or have a better sugestion.

**Current implementation (image and title centered, rest of text left aligned):**
image

**Other alignments I tried out:**
1) All centered:
image
2) All the text left aligned and image centered: 
image

## Checklist


lib/reports/utils.ts
```diff
@@ -3,8 +3,9 @@ import nodemailer from 'nodemailer'; // Nodemailer
 /**
  * Send email to user to confirm appliction was received
  * @param to Receiver's email
+ * @param firstName Receiver's first name
  */
-const sendConfirmationEmail = (to: string): Promise<void> => {
+const sendConfirmationEmail = (to: string, firstName?: string): Promise<void> => {
```
OustanDing, line 8: Why not make the `firstName` parameter non-nullable? Applications must have first names.

lib/reports/utils.ts
```diff
@@ -31,15 +31,14 @@ const sendConfirmationEmail = (to: string): Promise<void> => {
   });
 };
 
-// TODO: Update email content
-const text = 'Your parking permit application has been received';
+const text = 'Your Parking Permit Renewal Application was Successfully Submitted';
 
-// TODO: Update email content
-const html = `
+const html = (firstName?: string) => {
```
OustanDing, line 36: Same as above comment.

lib/reports/utils.ts
```diff
@@ -31,15 +31,14 @@ const sendConfirmationEmail = (to: string): Promise<void> => {
   });
 };
 
-// TODO: Update email content
-const text = 'Your parking permit application has been received';
+const text = 'Your Parking Permit Renewal Application was Successfully Submitted';
```
OustanDing, line 34: I want to clarify that the `text` parameter above is used as a fallback in case the HTML email cannot be rendered in the email client, so the `text` value should be very similar to, if not the same as the HTML content, just without the styling.

lib/reports/utils.ts
j-bunk: The issue starts in the `payment-received` file where TypeScript was yelling at me saying that `application` could be null which made me use the chain operator `?` which makes `firstName` nullable

image


I guess `application` can't be null in this context so I could use the non-null assertion Operator `const applicationFirstName = application!.firstName;` but I'd have to ignore the eslint rule

image

Lmk what you think!

lib/reports/utils.ts
j-bunk: Thanks for letting me know! Updated the text content accordingly

lib/reports/utils.ts
OustanDing: I see. This is an error in our code. Any `prisma.application.findUnique` call could return `null` (if the application ID doesn't exist). We should not have used optional chaining here - instead, we should've checked if `application` is `null`, and if it is, we should return a 400 (bad request).

Could you implement that in this PR? After that is fixed, we will be guaranteed that `application` is non-null when we try to access its properties.

j-bunk: Makes sense! Addressed!

lib/reports/utils.ts
j-bunk: Addressed!

pages/api/payment-received.ts
```diff
@@ -62,30 +62,36 @@ const paymentReceivedHandler: NextApiHandler = async (req, res) => {
...
-      data: {
-        shopifyPaymentStatus: ShopifyPaymentStatus.RECEIVED,
-        shopifyConfirmationNumber: `${shopifyOrderId}`,
-        paidThroughShopify: true,
-        donationAmount: donationAmount,
-      },
-    });
+    if (!application) {
+      // Bad request, occurs if application ID does not exist
+      res.status(400).end();
```
OustanDing, line 68: We should return an error message with the 400 request so that we know the reason for failure

OustanDing, line 68:
```suggestion
      return res.status(400).end();
```
and then the rest of the code doesn't need to live in an `else` block

j-bunk: Nice!!

j-bunk: Done! Lmk if you had a different error message in mind, currently have `application with given id does not exist` as the error message

OustanDing (approved):

lib/reports/utils.ts
```diff
@@ -74,11 +85,35 @@ const html = `
...
           </h1>
+          <p
+            style="
+              font-size: 18px;
+              line-height: 150%;
+              margin: 0 0 32px;
+            "
+          >
+          Hi ${firstName},
+          <br />
```
carelynntsai, line 98:
```suggestion
          <br /><br /> 
```

carelynntsai (approved):
Looks good! Can we just add an extra line break after the greeting?

j-bunk:
> Looks good! Can we just add an extra line break after the greeting?

Addressed! 
Final email: 
image

PR #145, opened 2022-02-26
[Misc] Form validation setup

angeladietz:
## Notion ticket link

[Set up form validation](https://www.notion.so/uwblueprintexecs/Set-up-form-validation-7a2ace2dcc97465bacd2a1090d69bc2d)

## Implementation description
* Installed Formik and Yup
* Added backend and frontend validation for the login page

Backend validation details:
* Added a Yup validation schema for login
* Throw one of the following error messages if login fails
  * `Invalid email input` if email does not match validation schema
  * `This email has not been registered by the admin.` if user does not exist
  * `Internal server error` for all other errors

Frontend validation details:
* Created a reusable form text field component that integrates with Formik error messages and field labels
* Refactored previous login form to use Formik components

https://user-images.githubusercontent.com/35577524/157576023-268faf5c-0c3c-4e2b-86ae-b9f28b6fc31d.mov

## Notes
* Currently we're putting backend errors into an alert right below the email input field. Designs for this are WIP so this implementation is temporary to unblock the ticket.
* When there is an AWS failure, we get back an `EmailSignin` error which is currently still shown (see screenshot). We could hardcode the name of this error to map it to a more useful string perhaps? I wouldn't expect to see this much in production, in this case it happened because i think I've added a@b.c as an email in the system so it thinks its valid and tries and fails to send an email to it

Screen Shot 2022-03-08 at 9 01 35 PM

## Checklist

### DIFF
--- a/components/form/TextField.tsx
+++ b/components/form/TextField.tsx
@@ -13,21 +13,31 @@ type Props = InputProps & {
   readonly name: string;
   readonly label: string;
   readonly required?: boolean;
+  readonly backendError?: string;
+  readonly setBackendError?: (updatedBackendError: string) => void;
 };
 
 const TextField: FC<Props> = props => {
-  const { name, label, required, ...inputProps } = props;
+  const { name, label, required, backendError, setBackendError, ...inputProps } = props;
   const [field, meta] = useField(name);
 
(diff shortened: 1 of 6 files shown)


OustanDing (approved):
LGTM

PR #147, opened 2022-03-03
[Feature] Make S3 Upload Links Private

EmilioMena:
## Notion ticket link

[Restrict view access for S3 documents](https://www.notion.so/uwblueprintexecs/Restrict-view-access-for-S3-documents-b8fbc6f6eebc476b997f77b0673ddd84)

## Implementation description
This PR makes the necessary changes to allow s3 files to be private while allowing access through pre-signed, temporary urls generated by our application. AWS changes documented in the ticket

Changes
- Create our own uploadToS3 function modified from the [next-s3-upload hook](https://github.com/ryanto/next-s3-upload/blob/master/packages/next-s3-upload/src/hooks/use-s3-upload.tsx)
  - main modification is changing the [ACL field](https://github.com/ryanto/next-s3-upload/blob/master/packages/next-s3-upload/src/hooks/use-s3-upload.tsx#L67) to be private
  - remove the hook and state features from it since we don't really need them
- Wrap the next-s3-upload API Route with our own auth to restrict unauthenticated clients from getting the temp federationToken key and thus being able to upload files

Presigned URL example
rcd8

## Notes
* High level overview of how `next-s3-upload` works. [Source](https://www.reddit.com/r/nextjs/comments/lmwhfx/best_practices_security_performance_for_uploading/gny6jrx/)

> you don't expose your AWS keys to the browser. You give them to the backend, and then the backend generates a pair of temporary/limited access keys so the browser can upload to the bucket.
> 
> One of the reasons I made this addon was to prevent people from doing any sort of misconfiguration mishaps where they accidentally expose their root keys. That would be bad news bears.
> 

## Checklist

### DIFF
--- a/lib/utils/upload-to-s3.ts
+++ b/lib/utils/upload-to-s3.ts
@@ -1,5 +1,7 @@
 import S3 from 'aws-sdk/clients/s3';
 
+// Adapted from https://github.com/ryanto/next-s3-upload/blob/master/packages/next-s3-upload/src/hooks/use-s3-upload.tsx
+
 export const uploadToS3 = async (file: File) => {
   const filename = encodeURIComponent(file.name);
   const res = await fetch(`/api/s3-upload?filename=${filename}`);
--- a/lib/utils/upload-to-s3.ts
+++ b/lib/utils/upload-to-s3.ts
@@ -1,9 +1,19 @@
 import S3 from 'aws-sdk/clients/s3';
 
+type UploadedObject = {
+  url: string;
+  bucket: string;
+  key: string;
+};
 // Adapted from https://github.com/ryanto/next-s3-upload/blob/master/packages/next-s3-upload/src/hooks/use-s3-upload.tsx
-
-export const uploadToS3 = async (file: File) => {
+/**
+ * Upload file to S3
+ * @param file File to upload
+ * @returns object with url, bucket and key
+ */
+export const uploadToS3 = async (file: File): Promise<UploadedObject> => {
   const filename = encodeURIComponent(file.name);
+  // call backend to generate temporary/limited s3 upload keys
   const res = await fetch(`/api/s3-upload?filename=${filename}`);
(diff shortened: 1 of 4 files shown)


lib/utils/upload-to-s3.ts
```diff
@@ -0,0 +1,47 @@
+import S3 from 'aws-sdk/clients/s3';
+
+type UploadedObject = {
+  url: string;
+  bucket: string;
+  key: string;
+};
+// Adapted from https://github.com/ryanto/next-s3-upload/blob/master/packages/next-s3-upload/src/hooks/use-s3-upload.tsx
```
OustanDing, line 8: Let's move this comment to be within the JSDoc comment below

pages/api/s3-upload.ts
```diff
@@ -1,4 +1,26 @@
...
+const uploadHandler = APIRoute.configure({});
+const jwtSecret = process.env.NA_JWT_SECRET;
+
+/**
+ * Auth wrapper for next-s3-upload API Route
+ * https://next-s3-upload.codingvalue.com
+ */
+const s3UploadHandler: NextApiHandler = async (req, res) => {
+  const token = await getToken({ req: req, secret: jwtSecret });
+  if (token) {
```
OustanDing, line 16: Instead of manually using the JWT to check for login, please use `getSession` from NextAuth, unless it's necessary to do this. You can see an example of the use in our GraphQL context configuration: https://github.com/uwblueprint/richmond-centre-for-disability/blob/staging/lib/graphql/context.ts

OustanDing (changes requested):

pages/api/s3-upload.ts
EmilioMena: 👍 oh nice didn't know we can do that

chrischan325 (approved):
Looks good to me!

OustanDing (approved):

PR #155, opened 2022-03-20
[Feature] Identity Verification Form Validation

angeladietz:
## Notion ticket link

[Identity Verification Form Validation](https://www.notion.so/uwblueprintexecs/Identity-Verification-Form-Validation-eea5f2fb1f6b4dab90e7dbef8f0380c5)

## Implementation description
* Wrap form with Formik form components
* replace all inputs with reusable form input fields to enable Formik validation
* Create Yup schema to validate identity input requirements

Unrelated to form changes:
* Changed verify identity backend logic to allow the renewal of expired permits. `verifyIdentity` resolver now uses the most recent permit instead of the applicant's active permit.
* Changed verify identity logic to prevent temporary permits from being renewed

### Demo
https://user-images.githubusercontent.com/51224641/159341873-49399b99-7c11-4240-a112-bf35fe3f527d.mov

## Notes
* UserID field only allows user to type in numbers (includes characters that can appear in numbers like `.`, `-`)
* phone number suffix field only allows a maximum of 4 characters to be entered. Any characters can be typed but an error will appear below it instructing users to only enter numbers
* The alert doesn't have a close button (doesn't seem to work when included, can look into it more if needed)

## Checklist


lib/applicants/verify-identity/validation.ts
```diff
@@ -0,0 +1,17 @@
...
+export const verifyIdentitySchema = object({
+  userId: number()
+    .typeError('User ID must be a positive number')
+    .required('Please enter your User ID')
+    .positive('User ID must be a positive number')
+    .integer('Please enter your User ID'),
+  phoneNumberSuffix: string()
+    .matches(/^\d+$/, 'Must only contain numbers')
+    .required('Please enter the last 4 digits of your phone number')
+    .min(3, 'Must be exactly 4 digits'),
```
j-bunk, line 15: I think this might be clearer  `.length(4, 'Must be exactly 4 digits'),`

angeladietz: yep I agree, just fixed

lib/applicants/resolvers.ts
```diff
@@ -448,7 +447,7 @@ export const verifyIdentity: Resolver<MutationVerifyIdentityArgs, VerifyIdentity
   if (activePermit === null) {
     return {
       ok: false,
-      failureReason: 'APP_DOES_NOT_EXPIRE_WITHIN_30_DAYS',
+      failureReason: 'APP_DOES_NOT_EXPIRE_WITHIN_30_DAYS', // NOTETOSELF: should this be a different reason
       applicantId: null,
     };
   }
```
j-bunk, line 453: @OustanDing should the failure reason for someone without an active permit also be `APP_DOES_NOT_EXPIRE_WITHIN_30_DAYS`. This would display the following message on the frontend: `Your current permit expiry date is too far away to request a renewal.
      Please check back when your permit is expiring within 30 days or less.` which I feel might be confusing/misleading

lib/applicants/verify-identity/validation.ts
```diff
@@ -0,0 +1,17 @@
+import { object, string, date, number } from 'yup';
+
+/**
+ * Verify identity validation schema
+ */
+export const verifyIdentitySchema = object({
+  userId: number()
+    .typeError('User ID must be a positive number')
+    .required('Please enter your User ID')
+    .positive('Please enter your User ID')
```
j-bunk, line 10: @angeladietz
The initial value of `userId` is 0 so the message `User ID must be a positive number` would pop up (since zero is considered a non-positive number) when the user doesn't fill in the field so I felt the `Please enter your User ID` might be clearer in this scenario. Lmk if you think there's a better way to address this!

image

lib/applicants/resolvers.ts
OustanDing: Resolved, cc @carelynntsai

lib/applicants/verify-identity/validation.ts
OustanDing: This sounds good to me

pages/verify-identity.tsx
```diff
@@ -96,74 +88,107 @@ export default function IdentityVerificationForm() {
...
+              validationSchema={verifyIdentitySchema}
+              onSubmit={handleSubmit}
+            >
+              {({ values }) => (
+                <Form style={{ width: '100%' }} noValidate>
+                  <Box marginBottom="48px" textAlign={'left'}>
+                    <NumberField
+                      name="userId"
+                      label="User ID number"
+                      required={true}
```
OustanDing, line 111: For future reference, for boolean props, you can just write the following to set it to `true`
```suggestion
                      required
```

pages/verify-identity.tsx
```diff
@@ -96,74 +88,107 @@ export default function IdentityVerificationForm() {
...
+                    </Link>
+                    <Button
+                      type="submit"
+                      variant="solid"
+                      isLoading={loading}
+                      disabled={
+                        loading ||
+                        !values.userId ||
+                        !values.phoneNumberSuffix ||
+                        !values.dateOfBirth
```
OustanDing, line 185: We should be able to use the `isValid` prop here: https://formik.org/docs/api/formik#isvalid-boolean

components/form/SelectField.tsx
```diff
@@ -20,7 +20,7 @@ const SelectField: FC<Props> = props => {
   const [field, meta] = useField(name);
 
   return (
-    <FormControl isInvalid={!!meta.error} isRequired={required}>
+    <FormControl isInvalid={!!meta.error && !!meta.touched} isRequired={required}>
```
OustanDing, line 23:
```suggestion
    <FormControl isInvalid={!!meta.error && meta.touched} isRequired={required}>
```

components/form/RadioGroupField.tsx
```diff
@@ -24,7 +24,7 @@ const RadioGroupField: FC<Props> = props => {
   };
 
   return (
-    <FormControl isInvalid={!!meta.error} isRequired={required}>
+    <FormControl isInvalid={!!meta.error && !!meta.touched} isRequired={required}>
```
OustanDing, line 27:
```suggestion
    <FormControl isInvalid={!!meta.error && meta.touched} isRequired={required}>
```

OustanDing (commented):
Looks great so far! Some logic changes needed, as discussed via Slack @j-bunk @carelynntsai

pages/verify-identity.tsx
angeladietz: `isValid` only returns false if there are errors, but there won't be any errors when the form hasn't been filled out yet, so the button will be initially enabled. In this case, clicking it will just show the errors for all fields. Would you prefer this to the current implementation?

j-bunk:
@OustanDing `verifyIdentity` now uses the most recent permit instead of the applicant's active permit to allow users to renew expired permits. I don't have the most context on this since it's not related to the form changes, but I tried to pattern match the `activePermit` implementation. Lmk how things look, everything is in this commit https://github.com/uwblueprint/richmond-centre-for-disability/pull/155/commits/46b0fde4947c02161b74fbfc2e08701d03af2b21

lib/applicants/utils.ts
```diff
@@ -21,3 +22,25 @@ export const getActivePermit = async (applicantId: number): Promise<Permit | nul
 
   return activePermits.length > 0 ? activePermits[0] : null;
 };
+
+/**
+ * Get the most recent permit of an applicant
+ * @param applicantId ID of the applicant
+ * @returns The most recent permit of the applicant
+ */
+export const getMostRecentPermit = async (applicantId: number): Promise<Permit> => {
```
j-bunk, line 31: Does every applicant have at least one permit? Trying to figure out if this can return null

lib/applicants/utils.ts
```diff
@@ -21,3 +22,25 @@ export const getActivePermit = async (applicantId: number): Promise<Permit | nul
...
+  const permit = await prisma.applicant
+    .findUnique({
+      where: { id: applicantId },
+    })
+    .permits({
+      orderBy: { createdAt: SortOrder.DESC },
+      take: 1,
+    });
+
+  if (!permit) {
```
OustanDing, line 41:
```suggestion
  if (permit.length === 0) {
```
Careful here, in JS, `[]` is truthy.

pages/verify-identity.tsx
OustanDing: Ah, that's a good point. Let's keep the current conditions and also disable the button if `isValid` returns false
```suggestion
                        // isValid is false ||
                        !values.userId ||
                        !values.phoneNumberSuffix ||
                        !values.dateOfBirth
```

OustanDing (approved):

carelynntsai (approved):
🤩

PR #156, opened 2022-03-20
[Feature] Created Additional Information Card + Update Review Information Modal

aaggarwal10:
## Notion ticket link

[Additional Information Card](https://www.notion.so/uwblueprintexecs/ac7b1f636a214a0cbf45f6b76f899b54?v=5414ee5fc5974f149921e192d6eb29b8&p=fc2f1645b07f4897a7bd5b2e3e5a7a5b)

## Implementation description
* Added Card Form only for viewing (no editing)
NEW (as of March 24):
Dashed borders:
image
Guardian information section (fixed address format):
image

OLD:
Card

## Notes
* Verify display states, what different fields should be showing on diff conditions. (i.e. usesAccessibleConvertedVan = false, should additional information even be visible)

## Checklist


components/admin/requests/processing/ReviewInformationModal.tsx
```diff
@@ -61,7 +62,10 @@ export default function ReviewInformationModalProps({
                 {requestType === 'REPLACEMENT' ? (
                   <ReasonForReplacementCard applicationId={applicationId} editDisabled />
                 ) : (
-                  <DoctorInformationCard applicationId={applicationId} editDisabled />
+                  <>
+                    <DoctorInformationCard applicationId={applicationId} editDisabled />
+                    <AdditionalInformationCard applicationId={applicationId} />
+                  </>
```
OustanDing, line 68: For new requests, there will also be a guardian/POA card and a physician assessment card. We will need to add an extra condition block here to account for those sections. We don't have a PhysicianAssessment card yet, but let's add the guardian card for now.

components/admin/requests/additional-questions/Card.tsx
```diff
@@ -0,0 +1,116 @@
...
+                  Reason
+                </Text>
+              </Box>
+              <Box>
+                <Text as="p" textStyle="body-regular" textAlign="left">
+                  {requiresWiderParkingSpaceReason
+                    ? requiresWiderParkingSpaceReason
+                    : otherRequiresWiderParkingSpaceReason
+                    ? otherRequiresWiderParkingSpaceReason
+                    : 'Not Given'}
```
OustanDing, line 105:
```suggestion
                  {requiresWiderParkingSpaceReason
                    || otherRequiresWiderParkingSpaceReason
                    || 'Not Given'}
```

OustanDing: Actually, the above code is not the exact code we should use. If `requiresWiderParkingSpaceReason` is non-null, we should convert the enum to a string, similar to lines 75-77 (loading method)

OustanDing (changes requested):
For the ReviewInformationModal, could you please remove the `<form>` wrapping around the cards? By having the form there, when we click the button, the form gets submitted and the page refreshes. Semantically, the cards are not a form (there are no inputs). Rather, remove the `type="submit"` from the button and just add `onClick={onConfirmed}` to the button.

components/admin/requests/processing/review-information.ts
```diff
@@ -0,0 +1,26 @@
...
+// Queries an Application by ID along with the associated permit, replacement, applicationProcessing, and applicant
+export const GET_REVIEW_INFORMATION_QUERY = gql`
+  query GetApplication($id: Int!) {
+    application(id: $id) {
+      id
+      type
+      applicant {
+        guardian
+      }
+    }
```
OustanDing, line 13: We don't want to query the guardian via the applicant, as that is already on record. We want to pull the guardian information filled out in the application itself. Instead, we want to query the following:
```suggestion
    application(id: $id) {
      __typename
      id
      type
      ... on NewApplication {
        guardianFirstName
        guardianMiddleName
        guardianLastName
        etc...
      }
    }
```
which uses [GraphQL fragments](https://graphql.org/learn/queries/#fragments) to only query the guardian fields when we are dealing with a new application

components/admin/requests/processing/review-information.ts
```diff
@@ -0,0 +1,26 @@
...
+
+// Get application request type
+export type GetReviewInformationRequest = QueryApplicationArgs;
+
+// Get application response type
+// TODO: Account for application types
+export type GetReviewInformationResponse = {
+  application: Pick<Application, 'id' | 'type'> & {
+    guardian: Pick<Applicant, 'guardian'>;
+  };
```
OustanDing, line 25: Types here need to be updated when above change is implemented.

OustanDing (changes requested):

OustanDing (approved):
Key changes made:
- Move GQL query from Review Modal level to Guardian Info card level (consistent with other cards, easier to reuse)
- Add `isSubsection` prop to cards when they are subsections of other cards - setting to `true` will make the border dashed (see designs)
- Small fix to GQL schema, fixed guardian fields being non-null when querying for new application

carelynntsai:
Hello! Just wanted to confirm that an "Address" subhead got added to the review doctor's information section since it wasn't in the old screenshot 
image

Also the below scenario shouldn't ever actually happen, this was just some test placeholder data right?
image

aaggarwal10:
First Design Comment Fixed in PR #156 
Second Design Comment Is from Old Process, No Longer Occurs with new logic

PR #161, opened 2022-03-28
[Feature] Server side invoice upload to S3

EmilioMena:
## Notion ticket link

[Server side invoice upload to S3](https://www.notion.so/uwblueprintexecs/Server-side-invoice-upload-to-S3)

rcd9

Edit: new pdf file name
image

## Implementation description
* Create s3-upload helper functions (serverSideUpload, clientSideUpload, getPresignedUrl)
* Upload generated pdf to s3
* Update field resolvers
* Update processing task card

## Notes
* TTL of the presigned invoice s3 link is set at the max of 7 days but that can be changed
* I opted to save the presigned the db and update it when querying if it has expired so we don't need to generate a new one on every read

## Checklist


lib/application-processing/resolvers.ts
```diff
@@ -874,15 +876,47 @@ export const updateApplicationProcessingGenerateInvoice: Resolver<
...
     application,
     session,
     // TODO: Remove typecast when backend guard is implemented
     application.applicationProcessing.appNumber as number,
     invoice.invoiceNumber
   );
 
-  return { ok: true };
+  // Upload pdf to s3
+  const s3InvoiceKey = `invoices/${randomUUID()}/${invoice.invoiceNumber}.pdf`;
```
EmilioMena, line 888: Not sure what the file path should be. Right not its just the invoice number but it can be changes. Also added the random uuid for dev purposes so theres less collisions

pages/admin/request/create-new.tsx
```diff
@@ -252,7 +252,7 @@ export default function CreateNew() {
     let poaFormS3ObjectKey = '';
     if (guardianPOAFile) {
       try {
-        const { key } = await uploadToS3(guardianPOAFile);
+        const { key } = await clientUploadToS3(guardianPOAFile, 'rcd/poaFiles');
```
EmilioMena, line 255: same here, not sure what the file path should be

prisma/schema.sql
```diff
@@ -226,7 +227,7 @@ CREATE TABLE application_processing (
   review_request_completed_employee_id INTEGER,
   review_request_completed_updated_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
   invoice_number INTEGER UNIQUE,
-  documents_url VARCHAR(255),
+  documents_url TEXT,
```
EmilioMena, line 230: added URL here so that we don't have to generate it every time we are displaying it. There's a backend check when querying to keep the temporary url fresh

lib/application-processing/field-resolvers.ts
```diff
@@ -0,0 +1,50 @@
...
+    return null;
+  }
+
+  // Update signed URL if it has expired
+  const DAY = 24 * 60 * 60 * 1000;
+  const daysDifference = new Date().getTime() - invoice.updatedAt.getTime();
+  if (daysDifference > DAY * 6) {
+    let signedUrl;
+    try {
+      signedUrl = getSignedUrlForS3(invoice.s3ObjectKey as string);
```
EmilioMena, line 29: I set the invoice number URL to be max of 7 days, but that can be customized by passing in the second parameter here.

EmilioMena, line 29: When fetching the invoice here, I refresh the presigned url if it's already expired. Did this so we don't have the generate a new url each time.

EmilioMena: This will also need to be changed if the ttl is modified when actually generating the invoice. Maybe we can set this as an env var?

lib/application-processing/resolvers.ts
```diff
@@ -874,15 +876,47 @@ export const updateApplicationProcessingGenerateInvoice: Resolver<
...
+  let uploadedPdf;
+  let signedUrl;
+  try {
+    uploadedPdf = await serverUploadToS3(pdfDoc, s3InvoiceKey);
+    signedUrl = getSignedUrlForS3(uploadedPdf.key);
+  } catch (error) {
+    throw new ApolloError(`Error uploading invoice pdf to AWS: ${error}`);
+  }
```
EmilioMena, line 896: set the ttl of the url to be max of 7 days by default. We can change this by passing in a duration param to this helper function

lib/invoices/utils.ts
```diff
@@ -86,10 +85,9 @@ export const generateApplicationInvoicePdf = (
     },
   });
 
-  const pdfDoc = printer.createPdfKitDocument(definition);
-  // TODO: Upload invoice to S3
-  pdfDoc.pipe(fs.createWriteStream('temp/invoice.pdf'));
-  pdfDoc.end();
+  const pdfDocument = printer.createPdfKitDocument(definition);
+  pdfDocument.end();
+  return pdfDocument;
```
EmilioMena, line 90: s3 sdk actually accepts a readable stream so we can skip saving the file to the server

components/admin/requests/processing/TasksCard.tsx
```diff
@@ -254,11 +253,34 @@ export default function ProcessingTasksCard({ applicationId }: ProcessingTasksCa
               disabled={!reviewRequestCompleted}
               color="black"
               onClick={handleGenerateInvoice}
+              isLoading={generateInvoiceLoading}
+              loadingText="Generate document"
+              fontWeight="normal"
+              fontSize="14px"
```
EmilioMena, line 259: not sure why but button was getting weird styling during loading but to explicitly specifying here seemed to do the trick

tools/admin/requests/view-request.ts
```diff
@@ -35,7 +39,7 @@ export type GetApplicationResponse = {
       | 'appNumber'
       | 'appHolepunched'
       | 'walletCardCreated'
-      | 'invoiceNumber'
+      | 'invoice'
```
OustanDing, line 42: We should pick the selected invoice fields (ie. do `& { invoice: Pick<Invoice, ...> }`)

tools/admin/requests/processing-tasks-card.ts
```diff
@@ -44,7 +48,7 @@ export type GetApplicationProcessingResponse = {
       | 'appNumber'
       | 'appHolepunched'
       | 'walletCardCreated'
-      | 'invoiceNumber'
```
OustanDing, line 47: Please pick invoice fields (see similar comment on `tools/admin/requests/view-request.ts`)

prisma/schema.sql
OustanDing: Are URLs not guaranteed to fit in 255 chars?

lib/application-processing/schema.ts
```diff
@@ -7,6 +7,7 @@ import { gql } from '@apollo/client';
 // TODO: Add nested resolver support for employees
 export default gql`
   type ApplicationProcessing {
+    id: Int!
```
OustanDing, line 10: We shouldn't need an `id` field here, ID should not be queryable on an application processing object

lib/utils/s3-utils.ts
```diff
@@ -0,0 +1,114 @@
...
+    };
+  }
+};
+
+/**
+ * Create a signed URL for the purpose of accessing s3 object.
+ * By default, bucket objects should be private so they will only be accessible with this link.
+ * ** NOTE: This should only be called on the server side for security purposes **
+ * @param objectKey
+ * @param duration How long the URL should be good for. If not specified will be max duration of 7 days
```
OustanDing, line 93:
```suggestion
 * @param duration How long the URL should be valid for (in seconds). If not specified, defaults to the max duration of 7 days
```

lib/utils/s3-utils.ts
```diff
@@ -0,0 +1,114 @@
...
+      secretAccessKey: process.env.S3_UPLOAD_SECRET as string,
+    },
+    region: process.env.S3_UPLOAD_REGION,
+  });
+
+  const MAX_DURATION = 604800; // 7 days
+  const params = {
+    Bucket: process.env.S3_UPLOAD_BUCKET,
+    Key: objectKey,
+    Expires: duration || MAX_DURATION,
```
OustanDing, line 109: If the duration is greater than 7 days, is it a valid input here?

pages/admin/request/create-new.tsx
OustanDing:
```suggestion
        const { key } = await clientUploadToS3(guardianPOAFile, 'rcd/guardian-forms');
```

pages/api/s3-upload.ts
```diff
@@ -2,9 +2,14 @@ import { APIRoute } from 'next-s3-upload';
...
 import { getSession } from 'next-auth/client';
 
-// Folder structure in s3 can be in the configure if needed
+// Configure folder structure
 // see https://next-s3-upload.codingvalue.com/s3-file-paths
-const uploadHandler = APIRoute.configure({});
+const uploadHandler = APIRoute.configure({
+  key(req, filename) {
+    const filePath = req.query.filePath as string;
+    return `${filePath}/${filename.replace(/\s/g, '-')}`;
```
OustanDing, line 10: Just want to double check: will this create any issues if our folder names have hyphens?

lib/application-processing/resolvers.ts
OustanDing:
```suggestion
  const s3InvoiceKey = `rcd/invoices/${randomUUID()}/${invoice.invoiceNumber}.pdf`;
```
Let's put all application-related files in a base folder called `rcd/`. The rest of this file path looks good!

lib/application-processing/resolvers.ts
OustanDing: Sounds good, could you please add a comment here for clarification (in case we need to change it later)?

lib/application-processing/resolvers.ts
```diff
@@ -874,15 +876,47 @@ export const updateApplicationProcessingGenerateInvoice: Resolver<
...
+      where: {
+        invoiceNumber: invoice.invoiceNumber,
+      },
+      data: {
+        s3ObjectKey: uploadedPdf.key,
+        s3ObjectUrl: signedUrl,
+      },
+    });
+  } catch {
+    // TODO: Error handling
```
OustanDing, line 910:
```suggestion
    throw new ApolloError("Error updating invoice metadata in DB");
```

lib/application-processing/field-resolvers.ts
OustanDing: Do you mean to set the TTL in days as an env var? I agree with that

components/admin/requests/processing/TasksCard.tsx
```diff
@@ -254,11 +253,34 @@ export default function ProcessingTasksCard({ applicationId }: ProcessingTasksCa
               disabled={!reviewRequestCompleted}
```
OustanDing, line 253: Let's also disable the button while the invoice generation is loading to prevent firing multiple times

components/admin/requests/processing/TasksCard.tsx
```diff
@@ -254,11 +253,34 @@ export default function ProcessingTasksCard({ applicationId }: ProcessingTasksCa
...
             >
               <Text textStyle="xsmall-medium">Generate document</Text>
             </Button>
-          ) : // TODO: Replace with link to download file
-          null}
+          ) : (
+            <Tooltip
+              hasArrow
+              closeOnClick={false}
+              label="Clicking on this link wil open the document in a new tab"
```
OustanDing, line 267:
```suggestion
              label="Clicking on this link will open the document in a new tab"
```

components/admin/requests/processing/TasksCard.tsx
```diff
@@ -254,11 +253,34 @@ export default function ProcessingTasksCard({ applicationId }: ProcessingTasksCa
...
+            >
+              <Link
+                href={invoice.s3ObjectUrl as string}
+                isExternal={true}
+                textStyle="caption"
+                textDecoration="underline"
+                padding="0px 16px"
+                color="primary"
+              >
+                {invoice.s3ObjectKey?.split('/').at(-1)}
```
OustanDing, line 280: Can we add a comment to make it clear that this is the file name?

OustanDing (changes requested):
Looks really good! I left a couple of comments

prisma/schema.sql
EmilioMena: yeah ran into this while testing. Presigned urls have lots of query params which causes it to go over

lib/application-processing/field-resolvers.ts
EmilioMena: Cool any idea as to what that should be? Rn it's max of 7 days

lib/utils/s3-utils.ts
EmilioMena: Good call, will add a check to cap it at 7 days

pages/api/s3-upload.ts
EmilioMena: should be okay since we are only doing the replace on the filename not the whole key?

OustanDing: True

lib/application-processing/schema.ts
EmilioMena: Added it initially because ts was mad in the invoice field resolver
image

But since it shouldn't be queryable, I just intersected an id type in the actual file. Lmk what u think

components/admin/requests/processing/TasksCard.tsx
EmilioMena: Ah good point, this is for uploadDocuments button tho. I added in the invoice generation button

lib/utils/format.ts
```diff
@@ -44,8 +44,11 @@ export const formatDate = (date: Date, dateInput = false): string => {
...
  * @returns {string} formatted date
  */
-export const formatDateYYYYMMDD = (date: Date): string => {
-  return new Date(date).toISOString().split('T')[0];
+export const formatDateYYYYMMDD = (d: Date): string => {
+  let date = new Date(d);
+  const offset = date.getTimezoneOffset();
+  date = new Date(date.getTime() - offset * 60 * 1000);
+  return date.toISOString().split('T')[0];
 };
```
EmilioMena, line 52: Noticed that this wasn't adjusting back to the right timezone. Converting to UTC would move the date forward or back by one at some points in the day
https://stackoverflow.com/questions/23593052/format-javascript-date-as-yyyy-mm-dd

.env.sample
```diff
@@ -29,4 +29,5 @@ S3_UPLOAD_REGION=""
 
 # Misc
 PROCESSING_FEE="26"
-CONFIRMATION_EMAIL_FROM=""
\ No newline at end of file
+CONFIRMATION_EMAIL_FROM=""
+INVOICE_LINK_DURATION_DAYS="7"
```
EmilioMena, line 33: env var for invoice link ttl. Made the corresponding changes when generating invoice and when fetching from db

lib/application-processing/resolvers.ts
```diff
@@ -874,15 +880,54 @@ export const updateApplicationProcessingGenerateInvoice: Resolver<
...
     invoice.invoiceNumber
   );
 
-  return { ok: true };
+  // Upload pdf to s3
+  // file name format: PP-Receipt-P<YYYYMMDD>-<invoice number>.pdf
+  const fileName = `PP-Receipt-P${formatDateYYYYMMDD(invoice.createdAt).split('-').join('')}-${
+    invoice.invoiceNumber
+  }.pdf`;
+  const s3InvoiceKey = `rcd/invoices/${fileName}`;
```
EmilioMena, line 896: The example invoice rcd gave us had this format so i thought i might as well use it. I think it's better than `<invoiceNumber>.pdf`

.env.sample
OustanDing:
```suggestion
INVOICE_LINK_TTL_DAYS="7"
```
Naming suggestion

lib/application-processing/field-resolvers.ts
```diff
@@ -20,10 +20,17 @@ export const applicationProcessingInvoiceResolver: FieldResolver<
     return null;
   }
 
-  // Update signed URL if it has expired
+  if (!process.env.INVOICE_LINK_DURATION_DAYS) {
```
OustanDing, line 23: Update env variable name here

lib/application-processing/resolvers.ts
```diff
@@ -885,12 +889,19 @@ export const updateApplicationProcessingGenerateInvoice: Resolver<
   );
 
   // Upload pdf to s3
-  const s3InvoiceKey = `invoices/${randomUUID()}/${invoice.invoiceNumber}.pdf`;
+  // file name format: PP-Receipt-P<YYYYMMDD>-<invoice number>.pdf
+  const fileName = `PP-Receipt-P${formatDateYYYYMMDD(invoice.createdAt).split('-').join('')}-${
+    invoice.invoiceNumber
+  }.pdf`;
```
OustanDing, line 895: Need to confirm that this will not add any extra whitespace characters when saving

lib/utils/format.ts
OustanDing: Great catch! This logic will likely get replaced when we do our upcoming date investigation, but for now it is a good solution

lib/application-processing/resolvers.ts
EmilioMena: replaced this so the linter doesn't format it like that

```javascript
const createdAtYYYMMDD = formatDateYYYYMMDD(invoice.createdAt).replace(/-/g, '');
const fileName = `PP-Receipt-P${createdAtYYYMMDD}-${invoice.invoiceNumber}.pdf`;
const s3InvoiceKey = `rcd/invoices/${fileName}`;
```

OustanDing (approved):

PR #162, opened 2022-03-29
[Feature] Hook Up Document Upload to Processing Tasks Card

chrischan325:
## Notion ticket link

[Hook Up Document Upload](https://www.notion.so/uwblueprintexecs/Hook-Up-Document-Upload-5032d186ba1a4c6fab3bd8f1a71cf059)

## Implementation description
* Created new TaskCardUploadField component similar to the PoaFormUploadField component
* Opens file explorer and can select pdf file as well as display the link to the pdf 
* Every time user selects new file, it will upload to S3, used useEffect to call handleSubmit function

Screen Shot 2022-03-28 at 9 40 40 PM

## Notes
* 

## Checklist


aaggarwal10 (approved):
Overall, lgtm. Just some small comments, might not even be problems. Wait for Oustan to have final check tmrw tho.

EmilioMena (changes requested):
Good work! Added some small comments and highlighted the changes that need to be made after merging the [invoice generation pr](#161)

There's also an issue with what is being saved in the DB. An object key != object url. Also since the bucket is private and no one is able to access the document with a raw link, we need to generate and save a [signed url](https://github.com/uwblueprint/richmond-centre-for-disability/blob/staging/lib/utils/s3-utils.ts#L96) to get access. I think we will need to do the following (not necessarily in this order):
1. Add a field for `documents_s3_object_key` in application-processing db schema 
2. make changes in `lib/application-processing/resolvers.ts` similar to those in [#161](https://github.com/uwblueprint/richmond-centre-for-disability/pull/161/files#diff-3ec7a3c1aaa269986a3a668ee4ff6b80473d30d50c40b0a1572913ad64626954) where we save both the object key and the generated signed object url in the database.
3. When fetching the the application-processing record in [applications/field-resolvers.ts](https://github.com/uwblueprint/richmond-centre-for-disability/blob/staging/lib/applications/field-resolvers.ts#L48), we re-generate the presigned documents url if it has expired already (see [applications-processing/field-resolver.ts](https://github.com/uwblueprint/richmond-centre-for-disability/pull/161/files#diff-9c155f66daffea146db6af2c97b8f0c0bbb876654e081fd94a02e1c742ca46e6R27-R54) for an example where this is done for the invoice url)
4. Determine a "time-to-live" a signed url for these documents and save this as an env variable to be used in the implementation of the above 2 points. For invoice urls this is This would entail adding an env variable for the "time-to-live" of a documents url `INVOICE_LINK_TTL_DAYS`

let's check in with @OustanDing for thoughts on the above

components/admin/requests/processing/TasksCard.tsx
```diff
@@ -105,6 +107,31 @@ export default function ProcessingTasksCard({ applicationId }: ProcessingTasksCa
...
+  const handleSubmit = async () => {
+    let documentS3ObjectKey = '';
+    if (uploadDocument) {
+      try {
+        const { key } = await uploadToS3(uploadDocument);
+        documentS3ObjectKey = key;
+      } catch (err) {
+        toast({
+          status: 'error',
+          description: `Failed to upload POA file: ${err}`,
```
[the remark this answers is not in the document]
OustanDing: Yeah, this error message is incorrect. Here, we are uploading a scanned copy of the wallet card, invoice, etc. Let's make the error message:
```suggestion
          description: `Failed to upload documents: ${err}`,
```

OustanDing:
Regarding @EmilioMena 's comment [above](https://github.com/uwblueprint/richmond-centre-for-disability/pull/162#pullrequestreview-926855114), we do need the pre-signed URLs, especially for these forms since they contain sensitive data, including scans of physical copies of renewal forms. We should implement the steps that @EmilioMena outlined, but with a much shorter TTL (1 hour) which gets generated on every view request. Only RCD will view these files. Thus, there will be a slight modification in that we don't store the signed URL in the DB, as it will get regenerated on every view.

Thanks for this catch @EmilioMena !

chrischan325:
> Good work! Added some small comments and highlighted the changes that need to be made after merging the [invoice generation pr](#161)
> 
> There's also an issue with what is being saved in the DB. An object key != object url. Also since the bucket is private and no one is able to access the document with a raw link, we need to generate and save a [signed url](https://github.com/uwblueprint/richmond-centre-for-disability/blob/staging/lib/utils/s3-utils.ts#L96) to get access. I think we will need to do a similar thing that was done for invoice generation (not necessarily in this order):
> 
> 1. Add a field for `documents_s3_object_key` in application-processing db schema
> 2. make changes in `lib/application-processing/resolvers.ts` similar to those in [#161](https://github.com/uwblueprint/richmond-centre-for-disability/pull/161/files#diff-3ec7a3c1aaa269986a3a668ee4ff6b80473d30d50c40b0a1572913ad64626954) where we save both the object key and the generated signed object url in the database.
> 3. When fetching the the application-processing record in [applications/field-resolvers.ts](https://github.com/uwblueprint/richmond-centre-for-disability/blob/staging/lib/applications/field-resolvers.ts#L48), we re-generate the presigned documents url if it has expired already (see [applications-processing/field-resolver.ts](https://github.com/uwblueprint/richmond-centre-for-disability/pull/161/files#diff-9c155f66daffea146db6af2c97b8f0c0bbb876654e081fd94a02e1c742ca46e6R27-R54) for an example where this is done for the invoice url)
> 4. Determine a "time-to-live" for the documents url and save this as an env variable to be used in the implementation of the above points. For invoice urls this is `INVOICE_LINK_TTL_DAYS`
> 
> let's check in with @OustanDing for thoughts on the above. We can also make this a follow up ticket if needed



> Good work! Added some small comments and highlighted the changes that need to be made after merging the [invoice generation pr](#161)
> 
> There's also an issue with what is being saved in the DB. An object key != object url. Also since the bucket is private and no one is able to access the document with a raw link, we need to generate and save a [signed url](https://github.com/uwblueprint/richmond-centre-for-disability/blob/staging/lib/utils/s3-utils.ts#L96) to get access. I think we will need to do a similar thing that was done for invoice generation (not necessarily in this order):
> 
> 1. Add a field for `documents_s3_object_key` in application-processing db schema
> 2. make changes in `lib/application-processing/resolvers.ts` similar to those in [#161](https://github.com/uwblueprint/richmond-centre-for-disability/pull/161/files#diff-3ec7a3c1aaa269986a3a668ee4ff6b80473d30d50c40b0a1572913ad64626954) where we save both the object key and the generated signed object url in the database.
> 3. When fetching the the application-processing record in [applications/field-resolvers.ts](https://github.com/uwblueprint/richmond-centre-for-disability/blob/staging/lib/applications/field-resolvers.ts#L48), we re-generate the presigned documents url if it has expired already (see [applications-processing/field-resolver.ts](https://github.com/uwblueprint/richmond-centre-for-disability/pull/161/files#diff-9c155f66daffea146db6af2c97b8f0c0bbb876654e081fd94a02e1c742ca46e6R27-R54) for an example where this is done for the invoice url)
> 4. Determine a "time-to-live" for the documents url and save this as an env variable to be used in the implementation of the above points. For invoice urls this is `INVOICE_LINK_TTL_DAYS`
> 
> let's check in with @OustanDing for thoughts on the above. We can also make this a follow up ticket if needed

components/admin/requests/processing/TasksCard.tsx
```diff
@@ -90,20 +93,50 @@ export default function ProcessingTasksCard({ applicationId }: ProcessingTasksCa
...
+      } catch (err) {
+        toast({
+          status: 'error',
+          description: `Failed to upload documents: ${err}`,
+          isClosable: true,
+        });
+        setApplicationDocuments(null);
+        return;
+      }
+    }
```
EmilioMena, line 124: couple things here. Can we:
- remove the state var and the if condition (file should always be present)
- rename the key variable when destructuring instead of creating a new variable

Could look something like
```ts
    try {
      const { key: documentS3ObjectKey } = await clientUploadToS3(applicationDoc, 'rcd/application-documents');
    } catch (err) {
      toast({
        status: 'error',
        description: `Failed to upload documents: ${err}`,
        isClosable: true,
      });
      return;
    }
```

components/admin/requests/processing/TaskCardUploadStep.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+    if (!uploadedFile) {
+      return;
+    }
+
+    setErrorMessage(null);
+
+    // Verify that uploaded file is < 5MB
+    if (uploadedFile && uploadedFile.size > FILE_SIZE_LIMIT) {
+      setErrorMessage('File exceeds maximum size');
```
EmilioMena, line 53: should this also return so that the onUploadFile function is not called for invalid files?

components/admin/requests/processing/TasksCard.tsx
```diff
@@ -90,20 +93,50 @@ export default function ProcessingTasksCard({ applicationId }: ProcessingTasksCa
...
+
+    await uploadDocuments({
+      variables: { input: { applicationId, documentsUrl: documentS3ObjectKey } },
```
EmilioMena, line 127: Since we won't be saving documentsUrl in the db
```suggestion
      variables: { input: { applicationId, documentsS3ObjectKey } },
```

components/admin/requests/processing/TasksCard.tsx
```diff
@@ -90,20 +93,50 @@ export default function ProcessingTasksCard({ applicationId }: ProcessingTasksCa
...
+    });
+    refetch();
+  };
+
+  const handleUndoDocumentsUpload = async () => {
+    setApplicationDocuments(null);
+    await uploadDocuments({
+      variables: { input: { applicationId, documentsUrl: null } },
```
EmilioMena, line 135: Since we won't be saving documentsUrl in the db
```suggestion
      variables: { input: { applicationId, documentsS3ObjectKey: null } },
```

lib/application-processing/resolvers.ts
```diff
@@ -947,6 +947,20 @@ export const updateApplicationProcessingUploadDocuments: Resolver<
...
+  let signedUrl;
+  try {
+    const durationSeconds = parseInt(process.env.APPLICATION_DOCUMENT_LINK_TTL_DAYS) * 24 * 60 * 60;
+    if (documentsUrl) {
+      signedUrl = getSignedUrlForS3(documentsUrl, durationSeconds);
+    }
+  } catch (e) {
+    throw new ApolloError(`Error uploading application document to AWS: ${e}`);
+  }
+
```
EmilioMena, line 963: If we're generating the signed URL on every query request, then we don't need to save it in the db, so these changes can be moved from here to `lib/applications/field-resolvers.ts`

lib/application-processing/resolvers.ts
```diff
@@ -956,7 +970,8 @@ export const updateApplicationProcessingUploadDocuments: Resolver<
       data: {
         applicationProcessing: {
           update: {
-            documentsUrl,
+            documentsUrl: signedUrl,
+            documentsS3ObjectKey: documentsUrl,
```
EmilioMena, line 974: The `documentsS3ObjectKey` should be added in the gql input schema and then saved here. `documentsUrl` will not exist on in the table after the schema changes so it can be removed

Ik it's a bit confusing but remember that S3ObjectKey != S3Url
```suggestion
            documentsS3ObjectKey,
```

lib/applications/field-resolvers.ts
```diff
@@ -43,11 +44,55 @@ export const applicationApplicantResolver: FieldResolver<
...
  * Fetch processing data of application
+ * Refresh temporary s3 application document url if it has expired.
  * @returns Application processing object
  */
 export const applicationProcessingResolver: FieldResolver<
   Application,
   Omit<ApplicationProcessing, 'invoice'>
 > = async (parent, _args, { prisma }) => {
-  return await prisma.application.findUnique({ where: { id: parent.id } }).applicationProcessing();
+  const application = await prisma.application
```
EmilioMena, line 54:
```suggestion
  const applicationProcessing = await prisma.application
```

lib/applications/field-resolvers.ts
```diff
@@ -43,11 +44,55 @@ export const applicationApplicantResolver: FieldResolver<
...
+    .findUnique({ where: { id: parent.id } })
+    .applicationProcessing();
+
+  if (!application) {
```
EmilioMena, line 58:
```suggestion
  if (!applicationProcessing) {
```

lib/applications/field-resolvers.ts
```diff
@@ -43,11 +44,55 @@ export const applicationApplicantResolver: FieldResolver<
...
+          },
+        },
+      });
+    } catch {
+      throw new ApolloError('Failed to update temporary application document URL');
+    }
+    return await prisma.application
+      .findUnique({ where: { id: parent.id } })
+      .applicationProcessing();
+  }
```
EmilioMena, line 96: Since we're not saving the url in the db, then we don't need to update it on every read. Instead we will generate the signed url from the objectKey and add it to the response. You can use the code that you have in `lib/application-processing/resolvers.ts` to generate the signedUrl

lib/applications/field-resolvers.ts
```diff
@@ -43,11 +44,55 @@ export const applicationApplicantResolver: FieldResolver<
...
+  return application;
```
EmilioMena, line 97: i think you should be able to do something like this to keep typescript happy

```suggestion
  return {
    ...application,
    documentsUrl: signedUrl
  };
```

prisma/schema.sql
```diff
@@ -228,6 +228,7 @@ CREATE TABLE application_processing (
   review_request_completed_updated_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
   invoice_number INTEGER UNIQUE,
   documents_url TEXT,
```
EmilioMena, line 230: we can remove the documents_url from the db schema since we will be generating a new one on every read

.env.sample
```diff
@@ -30,4 +30,5 @@ S3_UPLOAD_REGION=""
 # Misc
 PROCESSING_FEE="26"
 CONFIRMATION_EMAIL_FROM=""
-INVOICE_LINK_TTL_DAYS="7"
\ No newline at end of file
+INVOICE_LINK_TTL_DAYS="7"
+APPLICATION_DOCUMENT_LINK_TTL_DAYS="7"
```
EmilioMena, line 34: Oustan mentioned the TTL for documents should be 1 hour so the env var should probably be changed to 
```suggestion
APPLICATION_DOCUMENT_LINK_TTL_HOURS="1"
```

Then to convert into seconds for the signed url helper function:
```js
const durationSeconds = parseInt(process.env.APPLICATION_DOCUMENT_LINK_TTL_HOURS) * 60 * 60;
signedUrl = getSignedUrlForS3(documentsUrl, durationSeconds);
```

EmilioMena (changes requested):
Thanks for adding those changes in!

As @OustanDing pointed out, the signed url TTL will be super short (1 hour) for these documents so saving the url in the db doesn't make sense in this case and we should instead be generating a new signed url on every read. We should then only save the `documentsS3ObjectKey` in the db

I left some comments on what i think is missing to do this.

chrischan325:
> Thanks for adding those changes in!
> 
> As @OustanDing pointed out, the signed url TTL will be super short (1 hour) for these documents so saving the url in the db doesn't make sense in this case and we should instead be generating a new signed url on every read. We should then only save the `documentsS3ObjectKey` in the db
> 
> I left some comments on what i think is missing to do this.

So if we generate a new signed url on every read, do we still need documentsUrl in the schema at all? Or should it be removed?

components/admin/requests/processing/TasksCard.tsx
chrischan325: Here documentS3ObjectKey becomes out of scope for the uploadDocuments mutation that's outside the try catch. Should the mutation be moved within the try catch block as well ?

EmilioMena: ah i see good point. Let's just keep it the same as before then where `documentS3ObjectKey` is declared outside the try catch then assigned inside

lib/applications/field-resolvers.ts
chrischan325: I get a slight type error where string | undefined cannot be of type Maybe<String>. To fix this I let `documentsUrl: signedUrl || null`. Not sure if this is correct I am assuming documentsUrl can be null for when there is no document uploaded.

EmilioMena:
@chrischan325 Changes look great, thanks for adding those in 🔥 

I just committed a couple of small things:
1. nits on variable names, comments and a small change in `lib/applications/field-resolvers.ts`
2. error message was not showing up so i just moved the error message component outside the fileUrl conditional in `components/admin/requests/processing/TaskCardUploadStep.tsx`
3. added a tooltip on the link

EmilioMena (approved):

components/admin/requests/processing/TaskCardUploadStep.tsx
```diff
@@ -0,0 +1,132 @@
...
+                variant="ghost"
+                size={'xs'}
+                icon={<CloseIcon boxSize={'0.5em'} />}
+                onClick={onUndo}
+              />
+            </HStack>
+          </VStack>
+        </>
+      ) : (
+        <>
```
OustanDing, line 94: Don't need to wrap this in a fragment

components/admin/requests/processing/TaskCardUploadStep.tsx
```diff
@@ -0,0 +1,132 @@
...
+    if (uploadedFile && uploadedFile.size > FILE_SIZE_LIMIT) {
+      setErrorMessage('File exceeds maximum size');
+      return;
+    }
+    onUploadFile(uploadedFile);
+  };
+  return (
+    <>
+      {fileUrl ? (
+        <>
```
OustanDing, line 62: Don't need to wrap this in a fragment

components/admin/requests/processing/TaskCardUploadStep.tsx
```diff
@@ -0,0 +1,132 @@
...
+    }
+  };
+
+  /**
+   * Get file name from s3 URL
+   * Decode URL in case of file names with special characters
+   */
+  const getFileName = () => {
+    return fileUrl ? new URL(decodeURI(fileUrl)).pathname.split('/').at(-1) : null;
+  };
```
OustanDing, line 37: We actually don't need to wrap this in a function! Anything in the function body is run on every render, and since we are calling this function on every render of this component, we can just pull out the logic.
```suggestion
  /** File name retrieved from s3 URL */
  const fileName = fileUrl ? new URL(decodeURI(fileUrl)).pathname.split('/').at(-1) : null;
```

components/admin/requests/processing/TaskCardUploadStep.tsx
```diff
@@ -0,0 +1,132 @@
...
+                <Link
+                  href={fileUrl}
+                  isExternal={true}
+                  textStyle="caption"
+                  color="primary"
+                  textDecoration="underline"
+                >
+                  {getFileName()}
+                </Link>
+              </Tooltip>
```
OustanDing, line 82: Let's only render this `Tooltip` if `fileName` is not `null`

vedantdsn (approved):

OustanDing (approved):
LGTM! Great work!

PR #167, opened 2022-04-03
[Feature] Reason For Replacement Form Validation

angeladietz:
## Notion ticket link

[Reason For Replacement Form Validation](https://www.notion.so/uwblueprintexecs/Reason-For-Replacement-Form-Validation-2773eb887500494a97fb0cf86c8906d5)

## Implementation description
* 

## Notes
* Merge #165 before this one

## Checklist

### DIFF
--- a/lib/applications/validation.ts
+++ b/lib/applications/validation.ts
@@ -12,7 +12,7 @@ import { date, mixed, number, object, string } from 'yup';
 export const reasonForReplacementFormSchema = object({
   reason: mixed<ReasonForReplacement>()
     .oneOf(Object.values(ReasonForReplacement))
-    .required('Please seelct reason for replacement'),
+    .required('Please select reason for replacement'),
   lostTimestamp: date()
     .nullable()
     .default(null)
@@ -23,6 +23,7 @@ export const reasonForReplacementFormSchema = object({
           return new Date(originalValue);
         })
         .typeError('Please enter date APP was lost')
+        .max(new Date(), 'Date must be in the past')
         .required('Please enter date APP was lost'),
     }),
   lostLocation: string()
@@ -63,9 +64,9 @@ export const reasonForReplacementFormSchema = object({
 });
 
 /**
- * Nested reason for replacement form validation schema
+ * Validation schema for edit reason for replacement form
  */
-export const nestedReasonForReplacementFormSchema = object({
+export const editReasonForReplacementFormSchema = object({
   reasonForReplacement: reasonForReplacementFormSchema,
 });
 
--- a/lib/applications/validation.ts
+++ b/lib/applications/validation.ts
@@ -13,14 +13,18 @@ export const reasonForReplacementFormSchema = object({
   reason: mixed<ReasonForReplacement>()
     .oneOf(Object.values(ReasonForReplacement))
     .required('Please seelct reason for replacement'),
-  lostTimestamp: date().when('reason', {
-    is: 'LOST',
-    then: date()
-      .transform(originalValue => {
-        return new Date(originalValue);
-      })
-      .required('Please enter date APP was lost'),
-  }),
+  lostTimestamp: date()
+    .nullable()
+    .default(null)
+    .when('reason', {
+      is: 'LOST',
+      then: date()
+        .transform((_value, originalValue) => {
+          return new Date(originalValue);
+        })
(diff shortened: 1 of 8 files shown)


lib/applications/validation.ts
```diff
@@ -3,7 +3,71 @@ import {
...
+import { ReasonForReplacement } from '@prisma/client';
+import { date, mixed, number, object, string } from 'yup';
+
+/**
+ * Reason for replacement form validation schema
+ */
+export const reasonForReplacementFormSchema = object({
+  reason: mixed<ReasonForReplacement>()
+    .oneOf(Object.values(ReasonForReplacement))
+    .required('Please seelct reason for replacement'),
```
OustanDing, line 15:
```suggestion
    .required('Please select reason for replacement'),
```

lib/applications/validation.ts
```diff
@@ -3,7 +3,71 @@ import {
...
+  lostTimestamp: date()
```
OustanDing, line 16: Can we check that this is in the past?

lib/applications/validation.ts
```diff
@@ -3,7 +3,71 @@ import {
...
+    .default(null)
+    .when('reason', {
+      is: 'STOLEN',
+      then: number()
+        .typeError('Please enter police file number')
+        .required('Please enter police file number'),
+    }),
+  stolenJurisdiction: string().nullable().default(null).when('reason', {
+    is: 'STOLEN',
+    then: string().nullable(),
```
OustanDing, line 57: I believe this should be `required`

lib/applications/validation.ts
```diff
@@ -3,7 +3,71 @@ import {
...
+  stolenPoliceOfficerName: string().nullable().default(null).when('reason', {
+    is: 'STOLEN',
+    then: string().nullable(),
+  }),
+});
+
+/**
+ * Nested reason for replacement form validation schema
+ */
+export const nestedReasonForReplacementFormSchema = object({
```
OustanDing, line 68:
```suggestion
export const editRequestReasonForReplacementFormSchema = object({
```

lib/applications/validation.ts
```diff
@@ -3,7 +3,71 @@ import {
...
+    then: string().nullable(),
+  }),
+  stolenPoliceOfficerName: string().nullable().default(null).when('reason', {
+    is: 'STOLEN',
+    then: string().nullable(),
+  }),
+});
+
+/**
+ * Nested reason for replacement form validation schema
```
OustanDing, line 66: We should make it more clear that this schema is used in the form when editing a request

pages/admin/request/create-replacement.tsx
```diff
@@ -65,15 +65,15 @@ export default function CreateReplacement() {
   });
 
   /** Reason for replacement section */
-  const [reasonForReplacement, setReasonForReplacement] = useState<ReasonForReplacementFormData>({
+  const reasonForReplacement: ReasonForReplacementFormData = {
```
OustanDing, line 68: Can we rename this to `INITIAL_REASON_FOR_REPLACEMENT`? Also, we should move this object outside of this function component (it can still in this file, though) - that way, we avoid reinstantiating this object on every render.

OustanDing (changes requested):

lib/applications/validation.ts
angeladietz: It wasn't required before and it's optional in the designs, do you want this to be changed?

OustanDing: Ah, you're right. This is correct then.

pages/admin/request/create-replacement.tsx
```diff
@@ -43,6 +43,17 @@ import { PaymentType } from '@lib/graphql/types';
...
+/** Initial reason for replacement form values */
+const INITIAL_REASON_FOR_REPLACEMENT: ReasonForReplacementFormData = {
+  reason: null,
+  lostTimestamp: null,
+  lostLocation: null,
+  stolenJurisdiction: null,
+  stolenPoliceOfficerName: null,
+  stolenPoliceFileNumber: null,
+  eventDescription: null,
+};
```
OustanDing, line 55: Can we move this to the `create-replacement` tools folder?

OustanDing (approved):

PR #170, opened 2022-04-14
[Fix] Fix request and permit holder tables not updating after edit

OustanDing:
## Notion ticket link

[Ticket](https://www.notion.so/uwblueprintexecs/Fix-tables-not-updating-when-navigating-back-02f8c3aa3af04d4c8ee994a5b8655ce4)

## Implementation description
* Force refetch of requests/permit holders when navigating to table pages to include updates to individual records
* Implemented temporary fix for permit holders table to skip the cache (unideal, added alert comment)
* Fixed minor bug where there are 2 phone number fields in the edit permit holder information modal of the permit holder page

## Notes
* N/A

## Checklist

### DIFF
--- a/components/admin/permit-holders/permit-holder-information/EditModal.tsx
+++ b/components/admin/permit-holders/permit-holder-information/EditModal.tsx
@@ -198,18 +198,6 @@ export default function EditUserInformationModal({
                       {'Example: 000-000-0000'}
                     </FormHelperText>
                   </FormControl>
-
-                  <FormControl isRequired>
-                    <FormLabel>{'Phone number'}</FormLabel>
-                    <Input
-                      value={phoneNumber}
-                      onChange={event => setPhoneNumber(event.target.value)}
-                      type="tel"
-                    />
-                    <FormHelperText color="text.secondary">
-                      {'Example: 000-000-0000'}
-                    </FormHelperText>
-                  </FormControl>
                 </Stack>
               </Box>
 
--- a/pages/admin/permit-holders.tsx
+++ b/pages/admin/permit-holders.tsx
@@ -74,59 +74,65 @@ const PermitHolders: NextPage = () => {
   // This will avoid firing a query for each key the user presses
   const debouncedSearchFilter = useDebounce<string>(searchFilter, 500);
 
-  // Set page number to 0 after every filter or sort change
-  useEffect(() => {
-    setPageNumber(0);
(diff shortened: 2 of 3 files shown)


pages/admin/permit-holders.tsx
```diff
@@ -74,59 +74,65 @@ const PermitHolders: NextPage = () => {
...
-            addressLine1,
-            addressLine2,
-            city,
-            postalCode,
-            ...applicant
-          }) => ({
-            id,
-            name: {
+      // ! Temporary fix to force permit holder row updates when navigating (skip cache)
+      fetchPolicy: 'network-only',
```
EmilioMena, line 94: how come this is only needed in the permit-holders table?

EmilioMena: Also will this refetch the query after every filter change? If so this might get a bit expensive and we should probably think of a more efficient long term solution imo

i found this after doing a bit of research that may be helpful https://www.apollographql.com/blog/apollo-client/caching/when-to-use-refetch-queries/

components/admin/permit-holders/permit-holder-information/EditModal.tsx
```diff
@@ -198,18 +198,6 @@ export default function EditUserInformationModal({
...
-                    <FormLabel>{'Phone number'}</FormLabel>
-                    <Input
-                      value={phoneNumber}
-                      onChange={event => setPhoneNumber(event.target.value)}
-                      type="tel"
-                    />
-                    <FormHelperText color="text.secondary">
-                      {'Example: 000-000-0000'}
-                    </FormHelperText>
-                  </FormControl>
```
EmilioMena, line 212: 🔥

EmilioMena (approved):
lgtm! just left some concerns about making round trips to the server after every filter change.

pages/admin/permit-holders.tsx
EmilioMena: Side note: could we use/implement something like “nextFetchPolicy” where only on page load we bypass cache and on subsequent queries we don’t?

https://www.apollographql.com/docs/react/data/queries/#nextfetchpolicy

OustanDing: > how come this is only needed in the permit-holders table?
Not actually sure, I noticed that the permit holder table was not refetching as needed so I added this in as a temp fix.

As per our discussion today, we'll leave in this temporary fix for now in the interest of time, but I'll keep this comment  thread open and create a ticket to address this. Thanks for the catch!

PR #174, opened 2022-04-27
[Bugfix] Disallow Inactive Employee Login

chrischan325:
## Notion ticket link

[Disallow inactive employee login](https://www.notion.so/uwblueprintexecs/Disallow-inactive-employee-login-5adf3b5980fe46e88972063dca9946f0)

## Implementation description
* Active field is not a findUnique input so I used to select to grab the active field from the matched record and checked if the employee was active
* Because this operation is in a try catch block, the stored result `employee` is not accessible to the catch block so the conditional check if employee is active as well as the throw error statement is in the try block

## Notes
* 

## Checklist


OustanDing (approved):
LGTM!

PR #185, opened 2022-05-04
[Feature] Doctor Information Form Validation Refactoring

bonnie-chin:
## Notion ticket link

[Doctor Information Form Validation](https://www.notion.so/uwblueprintexecs/Doctor-s-Information-Form-Validation-c0c4fa58b797494cbca822fd37967a35)

## Implementation description
* modified db + graphql schemas to redefine mspNumber as a string
* Created Yup validation schema
* Refactored form fields with Formik wrapper and input fields
* Added validation schema and initial values to Formik wrappers in forms

image
image

## Tweaks Remaining:
- [ ] Spacing fixes between some of the form fields
- [ ] move optional text to beside form field

## Checklist


tools/admin/requests/doctor-information.ts
```diff
@@ -11,7 +11,7 @@ import {
 export type DoctorFormData = {
   firstName: string;
   lastName: string;
-  mspNumber: number | null;
+  mspNumber: string | null;
```
angeladietz, line 14: I think it's no longer necessary for this to be possibly null so you can change the type of `mspNumber` to just `string`. If you could also search through the codebase and find anywhere that it was being initialized to `null`, and instead initialize it to `''` like the other fields.

tools/admin/permit-holders/doctor-information.ts
```diff
@@ -9,7 +9,7 @@ import {
 export type DoctorFormData = Pick<
   Physician,
   'firstName' | 'lastName' | 'phone' | 'addressLine1' | 'addressLine2' | 'city' | 'postalCode'
-> & { mspNumber: number | null };
+> & { mspNumber: string | null };
```
angeladietz, line 12:
```suggestion
| 'mspNumber' > ;
```

Same as previous comment about mspNumber type, we don't need this field to be nullable so we can take it directly from the `Physician` object.

components/admin/requests/doctor-information/EditModal.tsx
```diff
@@ -10,35 +10,30 @@ import {
...
 import DoctorInformationForm from '@components/admin/requests/doctor-information/Form';
 import { DoctorFormData } from '@tools/admin/requests/doctor-information'; // GraphQL types
+import { Form, Formik } from 'formik';
+import { editPhysicianInformationSchema } from '@lib/physicians/validation';
 
 type EditDoctorInformationModalProps = {
   children: ReactNode;
   readonly doctorInformation: DoctorFormData;
-  readonly onSave: (physicianData: DoctorFormData) => void; // Callback that accepts the inputs defined in this page
+  readonly onSave: (applicationData: any) => void;
```
angeladietz, line 22: Is there a reason the type of this prop was changed? I think we should probably keep it as `physicianData: DoctorFormData` unless i'm missing something

components/admin/requests/doctor-information/EditModal.tsx
```diff
@@ -10,35 +10,30 @@ import {
...
-   * Handle edit submission
-   */
-  const handleSubmit = (event: SyntheticEvent) => {
-    event.preventDefault();
-    onSave(doctorInformation);
+  const handleSubmit = (values: { doctorInformation: DoctorFormData }) => {
+    // TODO: Backend errors
+    onSave({
+      ...values.doctorInformation,
+    });
```
angeladietz, line 36:
```suggestion
    onSave(values.doctorInformation);
```

components/admin/requests/doctor-information/Form.tsx
```diff
@@ -86,49 +38,42 @@ export default function DoctorInformationForm({
...
-          </FormControl>
+          </TextField>
         </Stack>
       </Box>
     </>
   );
 }
+
+// TODO: fix spacing between fields
+// TODO: get it to be savable???
```
angeladietz, line 79: Are these TODO's resolved?

angeladietz (changes requested):
Looks great!

components/admin/requests/doctor-information/EditModal.tsx
bonnie-chin: hmm I guess no major rationale, I just didn't see a reason for it to be defined this way so I tried to keep it in line with how the rest of the edit modals were written

angeladietz: Okay, I think if we could at least bring back the type to be `DoctorFormData` since thats always the type that will be passed to the onSave so it just makes the code a bit safer.

components/admin/requests/doctor-information/Form.tsx
angeladietz: Can you please delete these TODO comments if they're outdated?

angeladietz (changes requested):
Thanks!! Just two very tiny things

components/admin/requests/doctor-information/Form.tsx
bonnie-chin: Yup fs my bad! I thought I made a commit getting rid of them but ig it didn't go through :/

angeladietz (approved):
Woo looks great 🚀

PR #190, opened 2022-05-06
[Feature] Applicant facing renewal form validation

angeladietz:
## Notion ticket link

[Applicant-Facing Renewal Form Validation](https://www.notion.so/uwblueprintexecs/Applicant-Facing-Renewal-Form-Validation-b1cfebc0af414e6fa54e72fa9503d74c)

## Implementation description
* 

## Notes
* 

## Checklist

### DIFF
--- a/lib/applications/validation.ts
+++ b/lib/applications/validation.ts
@@ -322,7 +322,7 @@ export const applicantFacingRenewalContactSchema = object().shape(
         then: string()
           .typeError('Please enter a valid email address')
           .email('Please enter a valid email address')
-          .required('Please enter a valid phone number'),
+          .required('Please enter a valid email address'),
       }),
     receiveEmailUpdates: bool().when('updatedContactInfo', {
       is: true,
@@ -385,7 +385,11 @@ export const applicantFacingRenewalDoctorSchema = object().shape({
     .default(null)
     .when('updatedDoctor', {
       is: true,
-      then: string().typeError('Please enter a postal code').required('Please enter a postal code'),
+      then: string()
+        .typeError('Please enter a postal code')
+        .required('Please enter a postal code')
+        .min(6, 'Please enter a valid postal code')
+        .max(7, 'Please enter a valid postal code'),
     }),
   doctorPhoneNumber: string()
     .nullable()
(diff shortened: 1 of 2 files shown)


lib/applications/validation.ts
```diff
@@ -259,3 +259,142 @@ export const replacementFormSchema = object({
...
+      .nullable()
+      .default(null)
+      .when(['updatedContactInfo', 'contactEmailAddress'], {
+        is: (updatedContactInfo: boolean, contactEmailAddress: string) => {
+          return updatedContactInfo && !contactEmailAddress;
+        },
+        then: string()
+          .typeError('Please enter a valid phone number')
+          .required('Please enter a valid phone number')
+          .max(12, 'Please enter a valid phone number in the format 555-555-5555'),
```
OustanDing, line 312: Nice!

pages/renew.tsx
```diff
@@ -167,6 +157,80 @@ export default function Renew() {
...
+    personalCity: string;
+    personalPostalCode: string;
+  }) => {
+    setUpdatedAddress(!!Number(values.updatedAddress));
+    if (values.updatedAddress) {
+      setPersonalAddressLine1(values.personalAddressLine1);
+      setPersonalAddressLine2(values.personalAddressLine2);
+      setPersonalCity(values.personalCity);
+      setPersonalPostalCode(values.personalPostalCode);
+    }
```
OustanDing, line 173: Should set these to `null` if `updatedAddress` is false? Applies to the other sections as well

pages/renew.tsx
```diff
@@ -574,13 +650,7 @@ export default function Renew() {
...
                 onClick={handleSubmit}
                 loading={loading}
-                disabled={
-                  !applicantId ||
-                  !certified ||
-                  invalidPersonalAddress ||
-                  invalidContact ||
-                  invalidDoctor
-                }
+                disabled={!applicantId || !certified}
```
OustanDing, line 653: Would it be possible to disable this button on any of the sections being invalid?

angeladietz: I think we don't need that check because you can't proceed/navigate between forms until the current form you are on is valid. This means that you won't be able to reach this step to submit the renewal unless all of the other forms are already validated. As well, it would be a non-trivial change that would require effort to achieve

OustanDing: Good point

PR #204, opened 2022-05-26
[Feature] Payment form backend validation

angeladietz:
## Notion ticket link

[Payment Information Backend Validation](https://www.notion.so/uwblueprintexecs/Payment-Information-Backend-Validation-374e1d1d86f94226a23c88d54d7d0e7d)

## Implementation description
* Returned response from payment information mutation to the edit modal so that it can display error and not close the modal if there is an error.
* Added an error field to return type from graphql mutation so the form validation error can be returned to the frontend.

Example of how a backend form validation error appears in the edit modal. Note that the error shown was manually triggered.

https://user-images.githubusercontent.com/35577524/170796209-65ad4712-a4b4-4900-b9a4-dea5db01deb4.mov

## Notes
* 

## Checklist

### DIFF
--- a/components/admin/requests/payment-information/EditModal.tsx
+++ b/components/admin/requests/payment-information/EditModal.tsx
@@ -10,9 +10,6 @@ import {
   useDisclosure,
   Text,
   Box,
-  Alert,
-  AlertIcon,
-  AlertTitle,
 } from '@chakra-ui/react'; // Chakra UI
 import {
   PaymentInformationFormData,
@@ -21,6 +18,7 @@ import {
 import PaymentDetailsForm from '@components/admin/requests/payment-information/Form';
 import { Form, Formik } from 'formik';
 import { editPaymentInformationSchema } from '@lib/applications/validation';
+import ValidationErrorAlert from '@components/form/ValidationErrorAlert';
 
 type EditPaymentDetailsModalProps = {
   readonly children: ReactNode;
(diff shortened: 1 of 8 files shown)


components/admin/requests/payment-information/EditModal.tsx
```diff
@@ -67,8 +85,14 @@ export default function EditPaymentDetailsModal({
                 <ModalBody paddingY="20px" paddingX="4px">
                   <PaymentDetailsForm paymentInformation={values.paymentInformation} />
                 </ModalBody>
+                {error && (
+                  <Alert status="error" mt="12px" pt="20px" pb="20px">
+                    <AlertIcon />
+                    <AlertTitle>{error}</AlertTitle>
+                  </Alert>
+                )}
```
OustanDing, line 93: Can we extract this to a reusable component in the form inputs folder?

lib/applications/resolvers.ts
```diff
@@ -1064,9 +1066,20 @@ export const updateApplicationPaymentInformation: Resolver<
...
   const { input } = args;
-  const { id, donationAmount, shippingPostalCode, billingPostalCode, ...data } = input;
+
+  try {
+    await paymentInformationMutationSchema.validate(input);
+  } catch (err) {
+    if (err instanceof ValidationError) {
+      return {
+        ok: false,
+        error: err.message,
```
OustanDing, line 1077: Double check if we need to also have `error: null` in the response object on success

OustanDing (changes requested):

lib/applications/resolvers.ts
angeladietz: Good catch, yeah i think you're right that its needed

OustanDing (approved):

PR #212, opened 2022-06-05
[Improvement] Loading State in Card Components

aaggarwal10:
## Notion ticket link

[Loading spinner for view permit holder and view request pages](https://www.notion.so/uwblueprintexecs/Loading-spinner-for-view-permit-holder-and-view-request-pages-2a82eb54cf334f22a88b4f7c01b223ee)

https://www.loom.com/share/6ae7afaeeb19445aa249c5c3d59d2830

## Implementation description
* Added loading state to card components
* Not sure if completely works though as some cards still do not appear to load. Ref. to loom video.

## Checklist

### DIFF
--- a/components/admin/LayoutCard.tsx
+++ b/components/admin/LayoutCard.tsx
@@ -38,7 +38,9 @@ export default function PermitHolderInfoCard(props: PermitHolderInfoCardProps) {
       borderRadius="8px"
       {...props}
     >
-      {!loading && (
+      {loading ? (
+        <LoadingSpinner />
+      ) : (
         <>
           <Flex
             w="100%"
@@ -62,9 +64,9 @@ export default function PermitHolderInfoCard(props: PermitHolderInfoCardProps) {
             {editModal}
           </Flex>
           {divider && <Divider mb="20px" />}
+          {children}
         </>
       )}
-      {loading ? <LoadingSpinner /> : <>{children}</>}
     </GridItem>
   );
 }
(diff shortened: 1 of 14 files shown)


components/admin/LayoutCard.tsx
```diff
@@ -34,24 +38,33 @@ export default function PermitHolderInfoCard(props: PermitHolderInfoCardProps) {
...
+                updated
+              </Text>
+            )}
+            <Spacer />
+            {editModal}
+          </Flex>
+          {divider && <Divider mb="20px" />}
+        </>
+      )}
+      {loading ? <LoadingSpinner /> : <>{children}</>}
```
OustanDing, line 67: Put `children` in the same React fragment as above and make it a single ternary operator

OustanDing (commented):
Just one small comment

OustanDing (approved):
LGTM! Great work!

carelynntsai:
Hi! Great job. 

Is there a way to prevent the "No search results" empty state from appearing for the permit holders table before they load? 

For the requests and permit holder pages, it looks a little glitchy when some of the cards appear before other ones and how the cards move around, is there a way to prevent this as well? (@OustanDing this might be out of scope for this ticket)

OustanDing:
@carelynntsai I agree with your comment on the cards shifting around. We could force all the loading spinners to be synced, but it will be a little hacky. What are your thoughts on only showing the loading spinners when the user updates a card? On initial load, we would use the current implementation, which is that the cards are not visible while loading.

carelynntsai:
@OustanDing the second solution sounds good to me

leogjhuang:
Closed stale issue.

PR #215, opened 2022-06-09
[Fix] Frontend form validation design feedback

angeladietz:
## Notion ticket link

[Finalize frontend form validation](https://www.notion.so/uwblueprintexecs/Finalize-frontend-form-validation-195501c70b094e9ea63736b1ecd36901)

## Implementation description
* Fixed bug where permit holder information on new requests couldn't be saved
* Improved validation of postal codes and phone numbers using regex matching
* No changes implemented for date inputs. I confirmed that the same issue of the dates being prepopulated in new forms is not a new problem as it exists on staging as well, and this problem does not occur when using Chrome.

## Checklist

### DIFF
--- a/lib/guardian/validation.ts
+++ b/lib/guardian/validation.ts
@@ -67,9 +67,11 @@ export const guardianInformationSchema = object({
       is: false,
       then: string()
         .typeError('Please enter a postal code')
-        .min(6, 'Must be a valid postal code')
-        .max(7, 'Must be a valid postal code')
-        .required('Please enter a postal code'),
+        .required('Please enter a postal code')
+        .matches(
+          /(^(?!.*[DFIOQU])[A-VXY][0-9][A-Z] ?[0-9][A-Z][0-9]$)/,
+          'Please enter a valid postal code'
+        ),
     }),
 });
 
--- a/lib/guardian/validation.ts
+++ b/lib/guardian/validation.ts
@@ -1,3 +1,4 @@
+import { phoneNumberRegex, postalCodeRegex } from '@lib/utils/validation';
 import { object, bool, string } from 'yup';
 /**
  * Entire Guardian/POA validation schema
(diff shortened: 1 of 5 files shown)


lib/guardian/validation.ts
```diff
@@ -67,9 +70,11 @@ export const guardianInformationSchema = object({
       is: false,
       then: string()
         .typeError('Please enter a postal code')
-        .min(6, 'Must be a valid postal code')
-        .max(7, 'Must be a valid postal code')
-        .required('Please enter a postal code'),
+        .required('Please enter a postal code')
+        .matches(
+          /(^(?!.*[DFIOQU])[A-VXY][0-9][A-Z] ?[0-9][A-Z][0-9]$)/,
+          'Please enter a valid postal code'
```
OustanDing, line 76: We should also specify the correct postal code format in the error message, as we're doing it for phone number

lib/physicians/validation.ts
```diff
@@ -81,14 +81,21 @@ export const requestPhysicianInformationSchema = object({
   mspNumber: string()
     .matches(/^\d+$/, 'Must only contain numbers')
     .required('Please enter the MSP number'),
-  phone: string().min(10, 'Must be a valid phone number').required('Please enter a phone number'),
+  phone: string()
+    .required('Please enter a phone number')
+    .matches(
+      /(^(\(?\d{3}\)?-?\d{3}-?\d{4}$))/,
```
OustanDing, line 87: I notice we use the same regex for phone number and postal code for everywhere. Let's put those regex patterns in a file `lib/validation/utils.ts` and import from there. That way, if any of the regexes need to be changed in the future, it's an easy fix.

OustanDing (commented):
Thank you for looking into these issues! A few small improvements we can make

OustanDing (changes requested):
I think diff is missing the `utils/validation.ts` file

OustanDing (approved):
LGTM! Great job :)

PR #216, opened 2022-06-12
[Feature] Add donations integration with Shopify

OustanDing:
## Notion ticket link

[Ticket](https://www.notion.so/uwblueprintexecs/Add-donations-to-Shopify-cart-backend-c2f8fa66c25f49c381525bc7f6808ce5)

## Implementation description
* Added Shopify products for various donation amounts (set in environment variables)
* Moved Shopify checkout initiation logic to serverside to hide logic and reduce client-side load - also provides some atomicity for creating the application and creating the Shopify checkout
  * Some environment variables no longer need to be exposed to the frontend - need to update accordingly
* Modify `payment-received` webhook to update donation amount and billing information in application record
  * The Shopify payload includes the billing address but not the shipping address. When the application record is initially created, we set both billing and shipping addresses to be the same as home address. Billing address is updated upon receiving the webhook request from Shopify
  * Removed tipping at checkout

## Notes
* Need to make frontend changes for donation amount in applicant-facing renewal form once form validation is done

## Checklist


pages/api/payment-received.ts
```diff
@@ -73,6 +120,20 @@ const paymentReceivedHandler: NextApiHandler = async (req, res) => {
...
+      ? {
+          billingFullName: rawBillingInformation.name,
+          billingAddressLine1: rawBillingInformation.address1,
+          billingAddressLine2: rawBillingInformation.address2,
+          billingCity: rawBillingInformation.city,
+          billingProvince: rawBillingInformation.province_code,
+          billingCountry: rawBillingInformation.country,
+          billingPostalCode: stripPostalCode(rawBillingInformation.zip),
+        }
+      : {};
```
EmilioMena, line 135: Do we need to update billing info? From previous discussions i though we wanted source of truth to come from RCD platform. If we do, should we also be updating shipping?

lib/applications/resolvers.ts
```diff
@@ -541,32 +546,24 @@ export const createExternalRenewalApplication: Resolver<
...
-        shippingCountry: applicant.country,
-        shippingPostalCode: stripPostalCode(applicant.postalCode),
-        // TODO: Replace billing info with Shopify checkout inputs
+        shippingFullName: null,
+        shippingAddressLine1: null,
+        shippingAddressLine2: null,
+        shippingCity: null,
+        shippingProvince: null,
+        shippingCountry: null,
+        shippingPostalCode: null,
```
EmilioMena, line 557: How come shipping address is changed to being all null here? I also see that it's not being updated from shopify webhook. When would shipping address data be populated?

lib/shopify/utils.ts
```diff
@@ -0,0 +1,121 @@
+import Client from 'shopify-buy';
+
+/** Possible donation amounts in dollars */
+export type DonationAmount = 0 | 5 | 10 | 25 | 50 | 75 | 100;
```
EmilioMena, line 4: Are we expecting the set of donation amounts to possibly change in the future? If yes, i believe rcd would have to also change this type along with `env.SHOPIFY_PERMIT_PRODUCT_ID` which isn't super obvious.

I think we should either make a note to document this somewhere or make the type more generic.

EmilioMena (commented):
Nice work! Refactoring the shopify client into a separate class and calling it from the BE was a great call 🚀 

Left a couple small comments/concerns but other than that looks good.

Also were you able to test this?

pages/api/payment-received.ts
OustanDing: Good point. I double-checked the Shopify webhook payload which actually has the billing info, so I think it's reasonable to use this billing address. RCD can actually modify any payment information on any application (they need to manually correct applicant mistakes), so I feel that we can update billing info here.

Regarding shipping info, that is not present in the Shopify form or the payload, so we default to the home address, which is what RCD expects.

lib/applications/resolvers.ts
OustanDing: Shipping address fields should all be `null` when shipping address is the same as home address. It won't get updated because the Shopify webhook payload does not contain any shipping info. We default the shipping address to be the same as home address.

lib/shopify/utils.ts
OustanDing: Good point. I'll add an extra note that the environment variable also needs to be changed.

lib/applications/resolvers.ts
OustanDing: It is okay to set the shipping address fields to `null` because the `flattenApplication` function in `lib/applications/utils.ts` populates billing/shipping address fields when they are the same as home address

EmilioMena (approved):
LGTM if this was tested 🚀

OustanDing:
Thank you for the review @EmilioMena ! This has been rigorously tested locally

PR #217, opened 2022-06-12
[Feature] Replacement Form Backend Validation

bonnie-chin:
## Notion ticket link

[Replacement Form Backend Validation](https://www.notion.so/uwblueprintexecs/ac7b1f636a214a0cbf45f6b76f899b54?v=5414ee5fc5974f149921e192d6eb29b8)

## Implementation description
* Modelled off of Create New Request and Renewal Request changes

image
image

## Notes
* Unable to test validation error alert locally because of status 400 errors (not specific to my form)

## Checklist

### DIFF
--- a/lib/applicants/validation.ts
+++ b/lib/applicants/validation.ts
@@ -23,7 +23,7 @@ export const requestPermitHolderInformationSchema = object({
       then: mixed<Gender>().oneOf(Object.values(Gender)).required('Please select a gender'),
       otherwise: mixed<Gender>().oneOf(Object.values(Gender)).optional(),
     }),
-  otherGender: string().optional(),
+  otherGender: string().nullable().default(null),
   email: string().email('Please enter a valid email address').nullable().default(null),
   phone: string()
     .required('Please enter a valid phone number')
--- a/lib/applicants/validation.ts
+++ b/lib/applicants/validation.ts
@@ -32,7 +32,10 @@ export const requestPermitHolderInformationSchema = object({
   phone: string()
     .required('Please enter a valid phone number')
     .matches(phoneNumberRegex, 'Please enter a valid phone number in the format 000-000-0000'),
-  receiveEmailUpdates: boolean().required(),
+  receiveEmailUpdates: boolean().when('type', {
+    is: 'NEW' || 'RENEWAL',
+    then: boolean().required(),
+  }),
   addressLine1: string().required('Please enter an address'),
   addressLine2: string().nullable().default(null),
   city: string().required('Please enter a city'),
--- a/lib/applicants/validation.ts
+++ b/lib/applicants/validation.ts
@@ -23,7 +23,7 @@ export const requestPermitHolderInformationSchema = object({
       then: mixed<Gender>().oneOf(Object.values(Gender)).required('Please select a gender'),
       otherwise: mixed<Gender>().oneOf(Object.values(Gender)).optional(),
     }),
-  otherGender: string().nullable().default(null),
+  otherGender: string().optional(),
   email: string().email('Please enter a valid email address').nullable().default(null),
   phone: string()
     .required('Please enter a valid phone number')
--- a/lib/applicants/validation.ts
+++ b/lib/applicants/validation.ts
@@ -33,7 +33,7 @@ export const requestPermitHolderInformationSchema = object({
     .required('Please enter a valid phone number')
     .matches(phoneNumberRegex, 'Please enter a valid phone number in the format 000-000-0000'),
   receiveEmailUpdates: boolean().when('type', {
-    is: 'NEW' || 'RENEWAL',
+    is: (type: ApplicationType) => type === 'NEW' || type === 'RENEWAL',
     then: boolean().required(),
   }),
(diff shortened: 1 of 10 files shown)


lib/applicants/validation.ts
```diff
@@ -27,7 +27,14 @@ export const requestPermitHolderInformationSchema = object({
   phone: string()
     .required('Please enter a valid phone number')
     .max(12, 'Please enter a valid phone number in the format 000-000-0000'),
-  receiveEmailUpdates: boolean().required(),
+  receiveEmailUpdates: boolean()
+    .nullable()
+    .default(null)
+    .when('type', {
+      is: 'NEW',
```
OustanDing, line 34: Should be required for both new and renewal applications

lib/applicants/validation.ts
```diff
@@ -27,7 +27,14 @@ export const requestPermitHolderInformationSchema = object({
...
+      then: boolean().required('Please enter your date of birth'),
```
OustanDing, line 35: Incorrect error message

lib/applications/resolvers.ts
```diff
@@ -857,27 +857,109 @@ export const createReplacementApplication: Resolver<
...
+    donationAmount,
+    shippingAddressSameAsHomeAddress,
+    shippingFullName,
+    shippingAddressLine1,
+    shippingAddressLine2,
+    shippingCity,
+    shippingProvince,
+    shippingCountry,
+    shippingPostalCode,
+    billingAddressSameAsHomeAddress,
```
OustanDing, line 887: I think missing `billingFullName`

lib/applications/resolvers.ts
```diff
@@ -857,27 +857,109 @@ export const createReplacementApplication: Resolver<
   MutationCreateReplacementApplicationArgs,
   CreateReplacementApplicationResult
 > = async (_, args, { prisma }) => {
-  // TODO: Validation
   const { input } = args;
+
   const {
```
OustanDing, line 862: To avoid having to manually pass in all these fields below, we can destructure fields from `input` two times:
```typescript
const {
  applicantId,
  ...originalfields,
  ...data
} = input
const {
  ...new fields that need to be validated
} = input
```

This way, the `prisma` call below should remain unchanged

lib/applications/resolvers.ts
```diff
@@ -897,11 +979,33 @@ export const createReplacementApplication: Resolver<
       data: {
         type: 'REPLACEMENT',
         processingFee: process.env.PROCESSING_FEE,
-        donationAmount: donationAmount || 0,
+        firstName,
```
OustanDing, line 982: See comment above to avoid all these changes in this `prisma` call

lib/applications/validation.ts
```diff
@@ -199,7 +199,7 @@ export const paymentInformationMutationSchema = paymentInformationSchema.shape({
 export const reasonForReplacementFormSchema = object({
   reason: mixed<ReasonForReplacement>()
     .oneOf(Object.values(ReasonForReplacement))
-    .required('Please select reason for replacement'),
+    .required('Please select reason for replacement'), // TODO: consider making this optional??
```
OustanDing, line 202: What is the reason for making this field optional?

OustanDing (changes requested):

lib/applications/resolvers.ts
bonnie-chin: super good catch tysm

lib/applicants/validation.ts
bonnie-chin: Just finished rebasing -- I noticed that this changed to just be `receiveEmailUpdates: boolean().required(),` given this came from staging I'm assuming it should be kept that way?

lib/applicants/validation.ts
bonnie-chin: See comment above (no error message was indicated in the rebased version, which makes sense considering it's a checkbox)

lib/applicants/validation.ts
OustanDing: Yes!

lib/applicants/validation.ts
```diff
@@ -23,12 +23,19 @@ export const requestPermitHolderInformationSchema = object({
...
+    then: string().nullable().default(null),
+    otherwise: string().optional(),
+  }),
   email: string().email('Please enter a valid email address').nullable().default(null),
   phone: string()
     .required('Please enter a valid phone number')
     .matches(phoneNumberRegex, 'Please enter a valid phone number in the format 000-000-0000'),
-  receiveEmailUpdates: boolean().required(),
+  receiveEmailUpdates: boolean().when('type', {
+    is: 'NEW' || 'RENEWAL',
```
OustanDing, line 36: Don't think this will work, see https://github.com/jquense/yup#schemawhenkeys-string--string-builder-object--values-any-schema--schema-schema

We would need to use a function here to check that the type is either new or renewal. Please refer to the alternative code in the second example from the link above

OustanDing: Fixed

OustanDing (approved):

PR #208, opened 2022-05-29
[Feature] Physician Assessment Backend Validation

j-bunk:
## Notion ticket link

[Physician Assessment Backend Validation](https://www.notion.so/uwblueprintexecs/Physician-Assessment-Backend-Validation-3cda56a1148249468140274d5488c1ae)

## Implementation description
* Closely follows the implementation of https://github.com/uwblueprint/richmond-centre-for-disability/pull/204

Example of how the backend form validation error is displayed (the error was manually triggered):

https://user-images.githubusercontent.com/51224641/170884638-7814dc88-29fd-4276-8e4d-b240808a18cb.mov

## Notes
* 

## Checklist
- [x] My PR name is descriptive, is in imperative tense and starts with one of the following: `[Feature]`,`[Improvement]` or `[Fix]`,
- [x] I have run the appropriate linter(s)
- [x] I have requested a review from the RCD team on GitHub, or specific people who are associated with this ticket

PR #223, opened 2022-06-18
[Feature] Reason For Replacement Form Backend Validation

j-bunk:
## Notion ticket link

[Reason For Replacement Form Backend Validation](https://www.notion.so/uwblueprintexecs/Reason-For-Replacement-Form-Backend-Validation-df8f9a30b387469f9ee8c277f328174d)

## Implementation description
* Closely follows the implementation of https://github.com/uwblueprint/richmond-centre-for-disability/pull/208

---

Made a few minor fixes that are unrelated to backend validation in https://github.com/uwblueprint/richmond-centre-for-disability/commit/416c83ce6704e2518c7cdd309932ac4e894bc84e

Bugs fixed: 
* Could not update reason to `STOLEN` if it was previously `LOST` because of a frontend validation bug 
image
* Police file number would not prepopulate if it was previously set
image

* When updating the reason for a replacement to a different reason, some of the old irrelevant values are still kept in the db (e.g if we update the reason from `LOST` to `STOLEN`,  the value for `lostTimeStamp` still remained in the db even though it's irrelevant now). Changed this to set irrelevant values to `null` when updating the reason for replacement.

## Demo
Backend error:

image

Example of how the backend form validation error is displayed (the error was manually triggered):

image

## Notes
* 

## Checklist

### DIFF
--- a/components/admin/requests/reason-for-replacement/Card.tsx
+++ b/components/admin/requests/reason-for-replacement/Card.tsx
@@ -35,7 +35,8 @@ export default function ReasonForReplacementCard(props: ReplacementProps) {
     UpdateReasonForReplacementResponse,
     UpdateReasonForReplacementRequest
   >(UPDATE_REASON_FOR_REPLACEMENT);
-  const handleSave = async (data: ReasonForReplacementFormData) => {
+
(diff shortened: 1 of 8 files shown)


OustanDing (approved):
LGTM! Nice catch on the bug fixes too 🚀

PR #278, opened 2023-02-16
[Fix] Fix time zone for requests and reports

leogjhuang:
## Notion ticket link

[Ticket](https://www.notion.so/uwblueprintexecs/Some-time-zones-are-still-EST-8984f93c38da4fccab00a9b510cade56)

## Implementation description
Add new formatter to display request and report dates in local time zone instead of UTC.

## Notes
* All formatted dates with an associated time (e.g., requests, reports) were replaced with the local formatter.
* All formatted dates without an associated time (e.g., date of birth) were left untouched.

## Checklist

### DIFF
--- a/lib/utils/date.ts
+++ b/lib/utils/date.ts
@@ -19,6 +19,17 @@ export const formatDate = (date: Date, dateInput = false): string => {
  */
 export const formatDateYYYYMMDD = (d: Date, withTime = false): string => {
   const formatString = withTime ? 'YYYY-MM-DD, hh:mm a' : 'YYYY-MM-DD';
+  return moment.utc(d).format(formatString);
+};
+
+/**
+ * Format date to be in YYYY-MM-DD format and in local time zone
+ * @param {Date} date date to be formatted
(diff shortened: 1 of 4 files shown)


lib/utils/date.ts
```diff
@@ -19,7 +19,7 @@ export const formatDate = (date: Date, dateInput = false): string => {
  */
 export const formatDateYYYYMMDD = (d: Date, withTime = false): string => {
   const formatString = withTime ? 'YYYY-MM-DD, hh:mm a' : 'YYYY-MM-DD';
-  return moment.utc(d).format(formatString);
+  return moment(d).format(formatString);
```
sherryhli, line 22: > Need to test if this change affects other parts of the application (particularly date of birth)

In the interest of time, can we define a separate function to be used in reports and requests table only (e.g. `formatDateYYYYMMDDLocalTimeZone`)? This way we won't affect the other call sites and can speed up testing and push this out ASAP.

sherryhli (approved):
LGTM

PR #297, opened 2023-05-31
[Fix] No guardian on delete

leogjhuang:
## Notion ticket link
[Unable to create a new permanent permit for an existing temporary permit user](https://www.notion.so/uwblueprintexecs/3-Unable-to-create-a-new-permanent-permit-for-an-existing-temporary-permit-user-a5904b840ed6437ab050b648305410f0)
[Unable to create a new temporary permit for an existing temporary permit user](https://www.notion.so/uwblueprintexecs/4-Unable-to-create-a-new-temporary-permit-for-an-existing-temporary-permit-user-ee5d44fa75f94672a5b669e3bca762ec)

## Implementation description
* Turns out the issue with creating permits for existing users does not have to do with permanent/temporary
* Bug is replicated when user has no guardian attached to existing permit and a new permit is created with no guardian
* Attempts to delete the guardian field when it does not exist
* Made changes to application processing resolver and error catching

## Notes
* N/A

## Checklist

### DIFF
--- a/lib/application-processing/resolvers.ts
+++ b/lib/application-processing/resolvers.ts
@@ -331,7 +331,11 @@ export const completeApplication: Resolver<
                     },
                   },
                 }
-              : {}),
+              : {
+                  guardian: {
+                    disconnect: true,
+                  },
+                }),
           },
         });
 
--- a/lib/application-processing/resolvers.ts
+++ b/lib/application-processing/resolvers.ts
@@ -321,7 +321,7 @@ export const completeApplication: Resolver<
                 },
               },
             },
-            // If guardian information given, upsert. Otherwise, delete (SET NULL)
+            // If guardian information given, upsert.
             ...(guardian
               ? {
                   guardian: {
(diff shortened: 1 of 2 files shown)


lib/application-processing/resolvers.ts
```diff
@@ -331,7 +331,7 @@ export const completeApplication: Resolver<
                     },
                   },
                 }
-              : { guardian: { delete: true } }),
+              : {}),
```
sherryhli, line 334: Should we still account for the case where there was previously a guardian, but the new application does not contain a guardian? There should still be a delete right?

leogjhuang: Addressed this case by adding disconnect.

sherryhli (approved):

lib/application-processing/resolvers.ts
sherryhli: From what I understand this removes the mapping between the permit and the guardian, but there will be an orphaned guardian in the DB right? Ideally that should be cleaned up too.

PR #301, opened 2023-06-25
Show individual steps of the application process

ChinemeremChigbo:
seems to fix it!

## Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/ac7b1f636a214a0cbf45f6b76f899b54?v=5414ee5fc5974f149921e192d6eb29b8)

## Implementation description
* The reason some task logs were showing and ome were not is because for most applications the application ID did not match the applicationProcessing ID 
image
This was not a problem for applicationProcessingInvoiceResolver which always used the prisma.applicationProcessing but was for all the rest. I think this should fix the issue but it might need some further testing to see if anything is broken.

## Notes
* 

## Checklist

### DIFF
--- a/lib/application-processing/field-resolvers.ts
+++ b/lib/application-processing/field-resolvers.ts
@@ -63,9 +63,8 @@ export const applicationProcessingAppNumberEmployeeResolver: FieldResolver<
   ApplicationProcessing & { id: number },
   Employee
 > = async (parent, _, { prisma }) => {
-  return await prisma.application
+  return await prisma.applicationProcessing
     .findUnique({ where: { id: parent.id } })
-    .applicationProcessing()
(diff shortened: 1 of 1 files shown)


sherryhli:
Good catch!

> I think this should fix the issue but it might need some further testing to see if anything is broken.

Could you do a quick code search to see if any of these resolvers are used for anything else besides the task log?

sherryhli (approved):

ChinemeremChigbo:
> Good catch!
> 
> > I think this should fix the issue but it might need some further testing to see if anything is broken.
> 
> Could you do a quick code search to see if any of these resolvers are used for anything else besides the task log?

The resolvers that are changed are only used in the TasksCard.tsx file.

PR #311, opened 2023-08-29
[Feature] Tax receipt

ChinemeremChigbo:
## Notion ticket link

[[17] Tax Receipt PDF](https://www.notion.so/uwblueprintexecs/17-Tax-Receipt-PDF-e1605ce9e3b24ff3b5665561d08c88f0?pvs=4)

## Implementation description
* @sherryhli I'm not sure if I fully understand what a schema change is. For instance, in this ticket, I initially wanted to have the invoice and donation be separate pdfs, however this would have required me to change the `model ApplicationProcessing` in the `\prisma\schema.prisma` file to include some donationInvoiceNumber or make each applicationinvoice be a tuple of invoices (one for donation and the other for the invoice) or something similar. I believe this is a schema change since it adds or removes something from the schema.prisma file? Because of this I instead conditionally have both the invoice and the donation receipt in one file. If the donation amount exceeds $20 the file will have both invoice and donation, otherwise it will only have the invoice.

## Notes
* Added some comments throughout this PR for which a second opinion would be appreciated.

https://github.com/uwblueprint/richmond-centre-for-disability/assets/24593519/d3c30f08-0531-42cd-9d6f-764706105f79

## Checklist


public/assets/signature.png
ChinemeremChigbo, line 1: I did not use the new file RCD sent since I think this one looks better.

public/assets/stamp.png
ChinemeremChigbo, line 1: I did not use the new file RCD sent since I think this one looks better.

lib/invoices/utils.ts
```diff
@@ -242,3 +242,296 @@ const pdfDefinition = (input: {
...
+      applicantName,
+      userNumber,
+      permitType,
+      receiptNumber,
+      dateIssued,
+      issuedBy,
+      paymentItems,
+      totalAmount,
+      address,
+    }).content.concat([
```
ChinemeremChigbo, line 402: This is the invoice PDF. If we have a donation, we include both the invoice and donation PDFs.

sherryhli (approved):
> @sherryhli I'm not sure if I fully understand what a schema change is. For instance, in this ticket, I initially wanted to have the invoice and donation be separate pdfs, however this would have required me to change the model ApplicationProcessing in the \prisma\schema.prisma file to include some donationInvoiceNumber or make each applicationinvoice be an tuple of invoices (one for donation and the other for the invoice) or something similar. I believe this is a schema change since it adds or removes something from the schema.prisma file? Because of this I instead conditionally have both the invoice and the donation receipt in one file. If the donation amount exceeds $20 the file will have both invoice and donation, otherwise it will only have the invoice.

Yeah, you're right - adding donationInvoiceNumber would be considered a schema change, we'd need to add a column to the DB. I think your approach is a good workaround!

---
This looks great!

Thanks for the detailed walkthrough video. Would be great if you can write up a short paragraph describing those behaviours. I think all 3 features going out in the next release would benefit from having some user documention.

PR #318, opened 2023-10-01
[Fix] Set other gender field on permit holders page

leogjhuang:
## Notion ticket link

[Cannot save edits to personal info section (name or gender) when processing new requests](https://www.notion.so/uwblueprintexecs/23-Cannot-save-edits-to-personal-info-section-name-or-gender-when-processing-new-requests-00a593131eb5499a8bb2e3fbbafec3bb)

## Implementation description
* `otherGender` is set to null when updating permit holder information for new applicants
* Quick fix is to set `otherGender` to an empty string on save

### Error
other-gender-null

## Notes
* Once schema can be updated to allow `otherGender` to be null, this quick fix can be removed.

## Checklist

### DIFF
--- a/components/admin/requests/permit-holder-information/Card.tsx
+++ b/components/admin/requests/permit-holder-information/Card.tsx
@@ -94,6 +94,8 @@ const Card: FC<Props> = props => {
       | null;
     if (type === 'NEW') {
       const validatedData = await permitHolderInformationSchema.validate(permitHolderData);
+      // TODO: Remove this once schema is updated
+      validatedData.otherGender = '';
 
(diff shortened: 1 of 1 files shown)


sherryhli (approved):
Approved, feel free to merge so we can get a fix out ASAP. However, I think the schema in this case refers to the validation schema rather than the DB schema so we should be able to implement the long-term fix now.

Problem is somewhere here:
https://github.com/uwblueprint/richmond-centre-for-disability/blob/97b859053a53602138d9eff114d1062f1bb31457/lib/applicants/validation.ts#L26-L30

PR #320, opened 2023-10-03
Don't default to permanent if falsy permitType

ChinemeremChigbo:
The permit type field should never be falsy, but at times javascript eludes understanding

## Notion ticket link
[[16] For new temporary permit holders, under their Past APPs section, it lists them as having a permanent permit despite their recent permit request being a temporary one](https://www.notion.so/uwblueprintexecs/16-For-new-temporary-permit-holders-under-their-Past-APPs-section-it-lists-them-as-having-a-perma-8e37e85eb41740cb93d17e72a153f432)

## Implementation description
* I have still been unable to replicate this bug on my end, however I think it may be worth asking rcd if this change has any effect on their end.

## Checklist

### DIFF
--- a/components/admin/permit-holders/app-history/Card/AppHistoryRecord.tsx
+++ b/components/admin/permit-holders/app-history/Card/AppHistoryRecord.tsx
@@ -33,7 +33,7 @@ const AppHistoryRecord: FC<Props> = ({ permit }) => {
             <Text as="h3" textStyle="heading">
               Permit {rcdPermitId}
             </Text>
-            <PermitTypeBadge variant={permitType || 'PERMANENT'} />
+            <PermitTypeBadge variant={permitType} />
           </HStack>
           <HStack spacing="20px">
             <Text as="p" textStyle="body-regular">


sherryhli:
> but at times JavaScript eludes understanding

😂

@ChinemeremChigbo one thought, I think this may be related to #273, another of the unreleased fixes. If permit type was previously not queried from the backend, then it would be defaulting to PERMANENT on the frontend

ChinemeremChigbo:
> > but at times JavaScript eludes understanding
> 
> 😂
> 
> @ChinemeremChigbo one thought, I think this may be related to #273, another of the unreleased fixes. If permit type was previously not queried from the backend, then it would be defaulting to PERMANENT on the frontend

Ahh interesting, should I close this PR then? I don't think that this permitType should ever be falsy, so I don't really think this change will make any difference.
image

sherryhli:
@ChinemeremChigbo Yeah let's close it. It seems #273 was indeed the fix based on RCD's latest email.

PR #325, opened 2023-10-24
[Fix] Make replacement requests clearer

ChinemeremChigbo:
## Notion ticket link

[[24] Make replacement requests clearer](https://www.notion.so/uwblueprintexecs/24-Make-replacement-requests-clearer-f8b6511d86584315beea24380247fab0)

## Implementation description
* Renamed the reason of replacement from "others" to "mail lost".
* Set a default permit fee of $0 without any specified payment method.

image

## Checklist
- [ x ] My PR name is descriptive, is in imperative tense and starts with one of the following: `[Feature]`,`[Improvement]` or `[Fix]`,
- [ x ] I have run the appropriate linter(s)
- [ x ] I have requested a review from the RCD team on GitHub, or specific people who are associated with this ticket

### DIFF
--- a/pages/admin/request/create-replacement.tsx
+++ b/pages/admin/request/create-replacement.tsx
@@ -69,6 +69,8 @@ export default function CreateReplacement() {
 
   const toast = useToast();
   const router = useRouter();
+  const ZERO_REPLACEMENT_FEE_PAYMENT_DETAILS = INITIAL_PAYMENT_DETAILS;
+  ZERO_REPLACEMENT_FEE_PAYMENT_DETAILS.processingFee = '0';
 
   // Get applicant autofill information
   const [getApplicant] = useLazyQuery<GetSelectedApplicantResponse, GetSelectedApplicantRequest>(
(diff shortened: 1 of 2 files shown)


leogjhuang (approved):

pages/admin/request/create-replacement.tsx
```diff
@@ -219,7 +221,7 @@ export default function CreateReplacement() {
                 type: 'REPLACEMENT',
                 receiveEmailUpdates: false,
               },
-              paymentInformation: INITIAL_PAYMENT_DETAILS,
+              paymentInformation: ZERO_REPLACEMENT_FEE_PAYMENT_DETAILS,
```
sherryhli, line 224: Is it $0 replacement fee for all reasons or only for mail lost?

ChinemeremChigbo: I would assume only for mail lost (since that would not be the fault of the client) but something like "Lost" probably would be. The RCD request though isn't super clear, and we could clarify? What do you think?

The request from RCD was as follows "how a default permit fee of $0 without any specified payment method."

sherryhli: As discussed in meeting, I think current implementation sets $0 for all reasons, can you please double check? :)

ChinemeremChigbo: Yes it is. Now with the most recent changes the default replacement fee is changed back to $31 and becomes $0 if Mail Lost is selected as the reason for replacement. It stays goes back to $31 if either Lost or Stolen is selected as a reason. What do you think? 

https://github.com/uwblueprint/richmond-centre-for-disability/assets/24593519/e81389cc-1acb-460f-9eff-7d0de649a992

sherryhli (approved):

PR #327, opened 2023-10-24
[Fix] Fix validation schema for other gender field

leogjhuang:
## Notion ticket link

[Fix otherGender form validation](https://www.notion.so/uwblueprintexecs/Fix-otherGender-form-validation-4fd4026dbff54764b1d297485613793c)

## Implementation description
* Follow up to #318

## Checklist

### DIFF
--- a/lib/applicants/validation.ts
+++ b/lib/applicants/validation.ts
@@ -23,11 +23,7 @@ export const requestPermitHolderInformationSchema = object({
       then: mixed<Gender>().oneOf(Object.values(Gender)).required('Please select a gender'),
       otherwise: mixed<Gender>().oneOf(Object.values(Gender)).optional(),
     }),
-  otherGender: string().when('type', {
-    is: 'NEW',
-    then: string().nullable().default(null),
-    otherwise: string().optional(),
-  }),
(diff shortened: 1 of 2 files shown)


lib/applicants/validation.ts
```diff
@@ -23,11 +23,7 @@ export const requestPermitHolderInformationSchema = object({
       then: mixed<Gender>().oneOf(Object.values(Gender)).required('Please select a gender'),
       otherwise: mixed<Gender>().oneOf(Object.values(Gender)).optional(),
     }),
-  otherGender: string().when('type', {
-    is: 'NEW',
-    then: string().nullable().default(null),
```
sherryhli, line 28: I'm curious why this didn't work? When the request type is NEW, isn't the logic exactly the same as the new version?

sherryhli (approved):

lib/applicants/validation.ts
leogjhuang: I checked that the type is indeed 'NEW' for new applicants, so unless I am misinterpreting how the `yup` API works, I'm not sure where the issue is.

PR #328, opened 2023-10-24
[Fix] Adjust schema to evaluate date at validation time

leogjhuang:
## Notion ticket link

[Physician certification date](https://www.notion.so/uwblueprintexecs/25-Physician-certification-date-12294a50d9e045b3bb0e9a17dfaa9e9a)

## Implementation description
* Changes validation schema for physician certification date to dynamically compute current date with `new Date()` on each validation attempt

## Notes
* Unable to replicate issue exactly, can only assume this happens in a very rare instance where the request page is left open and not completed

## Checklist
- [x] My PR name is descriptive, is in imperative tense and starts with one of the following: `[Feature]`,`[Improvement]` or `[Fix]`,
- [x] I have run the appropriate linter(s)
- [x] I have requested a review from the RCD team on GitHub, or specific people who are associated with this ticket

PR #335, opened 2023-11-27
[Fix] Fix replacement request otherGender and lostTimestamp form validations

sherryhli:
## Notion ticket link

[[26] Replacement request date](https://www.notion.so/uwblueprintexecs/26-Replacement-request-date-05dd117cf6954da3ad303483bbb16a99)

## Implementation description
Contains 2 fixes, please review the commits individually:
* 7a47b9572cfcecd9002a5b0f5e2cf6f665d3afe5: reverts validation schema change from #327 to fix the `Field \"otherGender\" is not defined by type \"CreateReplacementApplicationInput` error seen in staging
  * `otherGender` is not part of the GraphQL schema for replacement and renewal requests and thus should be marked optional in the validation
  * when marked as nullable with default null, a null `otherGender` field is passed for these types of requests and were getting rejected by the backend
  * the reason why `nullable().default(null)` did not seem to work for new requests prior to #327 is because the request type was not passed into the validation
 * 838f5b0ef5760d4a0fd4f2fa21a63a3831aaedce: fixes the timezone issue in the replacement request lost timestamp
   * `lostTimestamp` value stored in the frontend state is a Date object constructed from a YYYY-MM-DD string - *the date constructor defaults to UTC in this case*
   * This UTC timestamp is stored in the database
   * On the read path, the value is displayed in local time
   * For example, say 2023-11-26 is inputted, which is intended to be in PST. This gets stored as 2023-11-26 00:00 UTC in the database. When displayed on the UI, it shows as 2023-11-24 16:00 PST

To prevent such timezone issues in the future, we should review all reads and writes of dates in the application. The requirement is that DB values should always be in UTC, while client-side values should always be in local timezone. Ideally, we centralize conversion to UTC when writing from the client, and conversion to local time when reading from the DB.

(please excuse my computer being unresponsive)
edit-replacement-request
create-replacement-request

## Notes
* N/A

## Checklist

### DIFF
--- a/lib/applications/validation.ts
+++ b/lib/applications/validation.ts
@@ -8,12 +8,13 @@ import {
   requestPhysicianInformationSchema,
 } from '@lib/physicians/validation';
 import { PaymentType, Province, ReasonForReplacement, ShopifyPaymentStatus } from '@prisma/client';
-import { bool, date, mixed, number, object, string } from 'yup';
+import { bool, date, lazy, mixed, number, object, string } from 'yup';
 import {
   AccessibleConvertedVanLoadingMethod,
   RequiresWiderParkingSpaceReason,
 } from '@lib/graphql/types';
 import { monetaryValueRegex, phoneNumberRegex, postalCodeRegex } from '@lib/utils/validation';
+import moment from 'moment';
 
 /**
  * Additional Questions form validation schema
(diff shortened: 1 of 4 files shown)


lib/applications/validation.ts
```diff
@@ -215,13 +216,16 @@ export const reasonForReplacementFormSchema = object({
...
     .when('reason', {
       is: 'LOST',
-      then: date()
-        .transform((_value, originalValue) => {
-          return new Date(originalValue);
-        })
-        .typeError('Please enter date APP was lost')
-        .max(new Date(), 'Date must be in the past')
-        .required('Please enter date APP was lost'),
+      then: lazy(() =>
```
sherryhli, line 219: Addition of `lazy` is not directly related to this bug, preemptively adding it in case something like #328 comes up

leogjhuang (approved):
Good catch on the `otherGender` issue!

PR #338, opened 2023-12-27
[Feature] Add second payment method for permit requests

leogjhuang:
## Notion ticket link

[[27] Accept two separate payment options](https://www.notion.so/uwblueprintexecs/27-Accept-two-separate-payment-options-3e4e89edb570492b95fde95392baa18e)

## Implementation description
* This PR is longer than usual, so each commit has been appropriately named to help with reviewing. The commits cannot be run separately, but feel free to test on the branch and follow the thought process commit-by-commit.
* There are 3 main areas to test that I've listed below with demos.

### Payment details form
* Create a new request and verify that a second payment can be added/removed.
* Edit an existing request and verify that a second payment can be added/removed.
* Demo: https://github.com/uwblueprint/richmond-centre-for-disability/assets/77544794/6b06e8e1-6dcb-44a3-873b-fd619958c935

### Accountant report
* Generate an accountant report and verify that a request with two payment methods show up as two separate entries.
* This should occur regardless of whether the two payment methods are the same (e.g., cash and cash).
* Demo: https://github.com/uwblueprint/richmond-centre-for-disability/assets/77544794/5596c6fb-4f63-4b92-8ea4-f3496438bc93

### Invoice and donation receipt
* Approve a request with two payment methods and generate the invoice and donation receipt.
* Verify that the processing fee and donation amount show up as separate entries for each payment method.
* Verify that the donation receipt is only generated when the sum of the donation amounts is at least $20.
* Demo: https://github.com/uwblueprint/richmond-centre-for-disability/assets/77544794/8f9bdbe6-95d2-4d41-8bde-ebd2b1d02eb3

## Notes
Ran into a couple issues during my testing that might not be related to my changes. Not urgent but require further testing.
* `TypeError: null is not an object (evaluating '_applications.result')` occurs after running `npx prisma migrate reset` without running `npx prisma db seed --preview-feature` afterwards.
* Accountant reports with the same start/end date cannot be redownloaded (i.e., clicking `Export as .CSV` with the same dates more than once results in an error).

## Checklist


prisma/schema.prisma
```diff
@@ -58,6 +58,10 @@ model Application {
   paymentMethod                    PaymentType             @map("payment_method")
   processingFee                    Decimal                 @map("processing_fee") @db.Money
   donationAmount                   Decimal                 @default(0.00) @map("donation_amount") @db.Money
+  paymentMethod2                   PaymentType?            @map("payment_method2")
+  processingFee2                   Decimal?                @map("processing_fee2") @db.Money
+  donationAmount2                  Decimal?                @map("donation_amount2") @db.Money
+  hasSecondPaymentMethod           Boolean                 @default(false) @map("has_second_payment_method")
```
sherryhli, line 64: This is different from what I envisioned, but it's definitely a much simpler approach and works well enough. No objections to going ahead with this for now, but we will need to re-evaluate if additional payment methods are requested in the future.

For extensibility, I was originally thinking we should make application : payment a 1 to many relationship by setting up a new payment_info table. This allows an arbitrary number of payment methods to be specified. However, it does involve a fairly substantial schema change and would be more tricky as we'd need to migrate the existing payment info over to the new table.

prisma/schema.prisma
```diff
@@ -58,6 +58,10 @@ model Application {
   paymentMethod                    PaymentType             @map("payment_method")
   processingFee                    Decimal                 @map("processing_fee") @db.Money
   donationAmount                   Decimal                 @default(0.00) @map("donation_amount") @db.Money
+  paymentMethod2                   PaymentType?            @map("payment_method2")
```
sherryhli, line 61: Nit about the naming: what about `second_payment_method`, `second_processing_fee`, etc.? `paymentMethod2` almost makes it seem like we're using it in favour over a deprecated `paymentMethod` column.

lib/applications/resolvers.ts
```diff
@@ -386,6 +395,9 @@ export const createNewApplication: Resolver<
       data: {
         type: 'NEW',
         donationAmount: donationAmount || 0,
+        paymentMethod2: hasSecondPaymentMethod ? paymentMethod2 : null,
+        processingFee2: hasSecondPaymentMethod ? processingFee2 || 0 : null,
+        donationAmount2: hasSecondPaymentMethod ? donationAmount2 || 0 : null,
```
sherryhli, line 400: Why is `hasSecondPaymentMethod` necessary? Do `paymentMethod2`, `processingFee2`, and `donationAmount2` default to null?

components/admin/requests/payment-information/Card.tsx
```diff
@@ -133,33 +141,64 @@ const Card: FC<Props> = props => {
...
+                ${donationAmount}
+              </Text>
+            </Box>
+            <Box>
+              <Text as="p" textStyle="caption" textAlign="left">
+                Paid with {titlecase(paymentMethod)}
+              </Text>
+            </Box>
+          </SimpleGrid>
+          {hasSecondPaymentMethod && (
```
sherryhli, line 172: Similar question: why not check if paymentMethod2 is null instead of using a special flag?

sherryhli (commented):
Thanks so much @leogjhuang  for working on this big feature and organizing your commits to be easily reviewable!

Aside from a few questions, this largely looks good to me! I'd get @ChinemeremChigbo to review this as well as he has more familiarity with the reports logic.

This does not affect production but I'm running into some issues when seeding, would you mind taking a look? Likely with this [application](https://github.com/uwblueprint/richmond-centre-for-disability/blob/75eb2f67a2beeaadb550080de764561dfe309d48/prisma/dev-seed-utils/applications.ts#L268)
```
Unknown arg `hasSecondPaymentMethod` in update.hasSecondPaymentMethod for type ApplicationUpdateInput. Did you mean `paymentMethod`? Available args:
```

prisma/schema.prisma
leogjhuang: I initially had the same idea of creating a one-to-many relationship but realized that the schema change would be a lot more complex and might unintentionally impact critical functionality. I agree that the suggested solution is more extensible, but since the few online merchants that take split payments will accept at most 2 methods anyway, I think this approach covers all bases.

prisma/schema.prisma
leogjhuang: Good point! Renamed variables to improve clarity [here](https://github.com/uwblueprint/richmond-centre-for-disability/pull/338/commits/8e6fb0df2f4e1bee1f13ba14f237397797680861)

lib/applications/resolvers.ts
leogjhuang: `hasSecondPaymentMethod` helps differentiate the following types of applications since `paymentMethod2` is null in both cases:

1. An application without a second payment method
2. An application where a user has selected that they would like to add a second payment method but has yet to choose which method

Although there is an approach that does not require `hasSecondPaymentMethod`, this solution significantly reduces complexity in `validation.ts` and maintains a similar schema design as the other application boolean properties such as `shippingAddressSameAsHomeAddress` and `billingAddressSameAsHomeAddress` that are linked to the setting of the shipping and billing address, respectively.

components/admin/requests/payment-information/Card.tsx
leogjhuang: See [comment](https://github.com/uwblueprint/richmond-centre-for-disability/pull/338#discussion_r1485737141)

sherryhli:
Almost forgot, were you able to take a look at the seeding issue?

leogjhuang:
> Almost forgot, were you able to take a look at the seeding issue?

@sherryhli Yup, I believe it was something to do with migrations! Let me know if there are still issues with seeding.

ChinemeremChigbo (approved):
Great work @leogjhuang!  The only thing I would be concerned about is that different payment methods might actually have different billing addresses. Not sure if that's something we should be concerned with (@sherryhli ?) . The `Accountant reports with the same start/end date cannot be redownloaded` does not seem to be related to your PR but seems pretty important to fix.

PR #371, opened 2024-08-19
[Fix] Force network call for accountant report download

sherryhli:
## Notion ticket link

[Accountant reports with the same start/end date cannot be redownloaded](https://www.notion.so/uwblueprintexecs/RCD-Accountant-reports-with-the-same-start-end-date-cannot-be-redownloaded-e5f09d8cb99e4f0784e38a187b69ecd0)

## Implementation description
S3 URL for the generated file expires after 10 seconds (second argument to `getSignedUrlForS3`). Since the Apollo client caches results of all GraphQL queries by default, repeated downloads with the same start and end dates return the same S3 URL, which would be expired. The solution is to force a network call when generating the accountant report in order to always get a fresh URL.
https://github.com/uwblueprint/richmond-centre-for-disability/blob/a9231a39bcf418bea6ebdb6b7a5498aa354ee098/lib/reports/resolvers.ts#L722

## Notes
Also added VS Code setting to use workspace version of Typescript

## Checklist

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,5 @@
 {
   "editor.defaultFormatter": "esbenp.prettier-vscode",
-  "editor.formatOnSave": true
+  "editor.formatOnSave": true,
+  "typescript.tsdk": "node_modules\\typescript\\lib"
 }
--- a/pages/admin/reports.tsx
+++ b/pages/admin/reports.tsx
@@ -25,6 +25,7 @@ export default function Reports() {
     GenerateAccountantReportResponse,
     GenerateAccountantReportRequest
   >(GENERATE_ACCOUNTANT_REPORT_QUERY, {
+    fetchPolicy: 'network-only',
     onCompleted: data => {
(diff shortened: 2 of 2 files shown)


.vscode/settings.json
```diff
@@ -1,4 +1,5 @@
 {
   "editor.defaultFormatter": "esbenp.prettier-vscode",
-  "editor.formatOnSave": true
+  "editor.formatOnSave": true,
+  "typescript.tsdk": "node_modules\\typescript\\lib"
```
ChinemeremChigbo, line 4: Not sure if this change should be committed

ChinemeremChigbo (approved):
Not sure if the change to the settings.json should be submitted. Other than that, LGTM.

.vscode/settings.json
sherryhli: It wasn't gitignored, so we should either do that or leave this in. Otherwise, I'd need to add this in locally everytime. I think this setting should be shared though, we should all use the workspace version of TypeScript.

ChinemeremChigbo: Alright, feel free to merge!

PR #372, opened 2024-08-31
[enhancement] Track application status to return to original tab in admin view

jennchenn:
## Notion ticket link

[[RCD] Resuming on the previous page](https://www.notion.so/uwblueprintexecs/RCD-Resuming-on-the-previous-page-9de56dfe2f4d4cbbb0b321d6fa04322f?pvs=4)

## Implementation description
* Pass application status in router query to keep track of which tab user originated from
* Return to original tab when user clicks `← All Requests`

## Notes
* Default when navigating to `/admin` is to show the `Pending` applications
* If the user navigates to a request specifically (i.e. without `?origin=`) and then clicks  `← All Requests`, it will bring them back to the same tab as the application's status
  * I couldn't bring the user back to the same tab as the application status in all scenarios due to the `All` tab

## Checklist


sherryhli (approved):
LGTM!

PR #389, opened 2025-01-15
Create Wallet Card

ssreekar:
## Notion ticket link

[Create Wallet Card](https://www.notion.so/uwblueprintexecs/RCD-Generate-wallet-cards-526b2764b3284dc9855e54040305d4ae)

## Implementation description
* Added s3 bucket for wallet generation
* Added New database to store Wallet S3 information
* Added code to push wallet to s3 on successful wallet generation
* Added PDF generation code for wallet
* Updated error handling when issues occur

## Notes
* Check if wallet if generated in all 3 locations
image
image
image

Video highlighting where the new wallet card is created
https://github.com/user-attachments/assets/c7d2b6db-1c37-4d8c-ac56-c5120eb9cc8c

## Checklist


lib/application-processing/field-resolvers.ts
```diff
@@ -56,6 +57,58 @@ export const applicationProcessingInvoiceResolver: FieldResolver<
...
+  if (!walletCard) {
+    return null;
+  }
+
+  // Update the signed S3 URL if it has expired.
+  let signedUrl;
+  try {
+    signedUrl = refreshS3SignedUrl(
+      walletCard.s3ObjectKey,
+      walletCard.s3ObjectUrl,
```
sherryhli, line 82: I know this is an existing pattern, but I don't think we need to store the original signed S3 object URL in the DB at all, just generate a new signed URL using the object key when a read request comes in.

Let's leave it like this for now, but a potential task to add to the backlog

lib/application-processing/field-resolvers.ts
```diff
@@ -56,6 +57,58 @@ export const applicationProcessingInvoiceResolver: FieldResolver<
...
+      walletCard.updatedAt.getTime()
+    );
+  } catch (error) {
+    const message = `Failed to get AWS wallet card URL: ${error}`;
+    logger.error({ error: message });
+    throw new ApolloError(message);
+  }
+  // If S3 Url was updated
+  if (walletCard.s3ObjectUrl !== signedUrl) {
```
sherryhli, line 91: With the proposed change in the future, this part won't be needed

lib/application-processing/resolvers.ts
```diff
@@ -646,6 +652,16 @@ export const completeApplication: Resolver<
     throw new ApolloError(`Error completing application with ID ${id}`);
   }
 
+  // Generate the Wallet Card
+  const { id: employeeId } = session;
+  const walletCardResult = await createWalletCard(id, true, employeeId, prisma, logger);
```
sherryhli, line 657: Should wallet card creation be part of the complete application transactions above?
https://github.com/uwblueprint/richmond-centre-for-disability/blob/staging/lib/application-processing/resolvers.ts#L532-L536

lib/applications/resolvers.ts
```diff
@@ -636,6 +636,7 @@ export const createRenewalApplication: Resolver<
         secondDonationAmount: hasSecondPaymentMethod ? secondDonationAmount || 0 : null,
         phone: stripPhoneNumber(phone),
         ...data,
+        processingFee: processingFee || 0,
```
sherryhli, line 639: Was this broken before? :o (or why is this change needed?)

lib/utils/s3-utils.ts
```diff
@@ -131,3 +131,25 @@ export const getSignedUrlForS3 = (
...
 };
+
+/**
+ * Update S3SignedUrl for expired S3 Object Key
+ * @param objectKey S3 object key
+ * @param s3SignedUrl Current S3 Signed Url (possibly expired)
+ * @param lastUpdatedTime Last updating time for Object Key (UTC)
+ * @returns New a valid S3 Signed Url for Object Key
+ */
+export const refreshS3SignedUrl = (
```
sherryhli, line 142: Ahh did not realize this was new util function until now!

I think we can maybe just go with the approach of always generating a new signed URL when we fetch a file. Generating the URL isn't that expensive of an operation and it seems like a very specific case to optimize for. Depending on frequently RCD views the file we might always be refreshing. WDYT? :)

sherryhli (approved):
This was a beast of a PR, thanks for working on this @ssreekar! 🔥

Functionality looks good! Left some suggestions to consider before merging:
1) Include wallet card DB update in the application completion transaction
2) S3 URL management
    a) Always generate new URL on fetch
    b) (Future improvement) Stop storing S3 URL in DB, only store object key
 
Tagging @ChinemeremChigbo for another set of eyes

lib/application-processing/field-resolvers.ts
ssreekar: Good point, I'll add a ticket to the RCD Task board, probably don't want to add too much to this already big PR

lib/application-processing/resolvers.ts
ssreekar: Yeah good point, I refactored some of the code so that it is now apart of the transaction, wherever possible. Although since interactive transactions aren't available in prisma 2.8 much of what can be done in this case is limited. The updated code is much safer tho (aka if the wallet card pdf generation fails the application still completes)

lib/utils/s3-utils.ts
ssreekar: Yeah makes sense, I refactored this functionality for cleaner code's sake but this makes this update much simpler

lib/application-processing/resolvers.ts
sherryhli: > if the wallet card pdf generation fails the application still completes

Probably also a follow up but it would make sense to allow wallet card generation to be manually triggered then

> Although since interactive transactions aren't available in prisma 2.8 much of what can be done in this case is limited

Yeah that's unfortunate :/
It's possible to have incomplete wallet card records then (employee and application processing id get persisted, but no S3 key). I guess that's a stronger argument to provide a way to retry the pdf generation

sherryhli (approved):

PR #398, opened 2025-07-13
Ss/fix wallet card

ssreekar:
## Notion ticket link

[Ticket Name](https://www.notion.so/uwblueprintexecs/ac7b1f636a214a0cbf45f6b76f899b54?v=5414ee5fc5974f149921e192d6eb29b8)

## Implementation description
* 

## Notes
* 

## Checklist

### DIFF
--- a/lib/application-processing/resolvers.ts
+++ b/lib/application-processing/resolvers.ts
@@ -400,7 +400,7 @@ export const completeApplication: Resolver<
             firstName,
             lastName,
             dateOfBirth,
-            id
+            applicantId.toString()
           );
           if (!result || !result.ok) {
             logger.error({ error: 'Error creating Wallet Card PDF' });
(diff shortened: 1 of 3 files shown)


lib/application-processing/resolvers.ts
```diff
@@ -767,7 +782,7 @@ export const completeApplication: Resolver<
             firstName,
             lastName,
             updatedApplicant.dateOfBirth,
-            id
+            id.toString()
```
ChinemeremChigbo: The replacement wallet cards should also use the applicant id for “User #” instead of the application id
```suggestion
            applicantId.toString()
```

ChinemeremChigbo (changes requested):

ChinemeremChigbo:
Looks good to me other than that change!

lib/application-processing/resolvers.ts
ssreekar: Great catch, updated!

PR #408, opened 2025-11-25
Add validation for max APP number in AssignNumberModal

ChinemeremChigbo:
## Implementation description
* Introduced a MAX_APP_NUMBER constant and updated validation logic to ensure the assigned number is an integer and does not exceed 2,147,483,647.
image
image


ssreekar (approved):
LGTM, one note to add is that it would be nice to have a min_APP_NUMBER as well, for -1 cases. Approving regardless as that might be out of scope for this ticket, leaving up to your disgression

